PR #3, opened 2026-07-02
Port build-registry to pure bash, drop the python dependency

JordanTheJet:
Follow-up to #2 (this commit was pushed to the branch minutes after the merge, so it missed the train).

Ports `tools/build-registry.py` to `tools/build-registry.sh` — same behavior, no python. Differentially tested against the python version on identical staged input:

- `registry.json` byte-identical (minus `sha256`, which differs only because each tool builds its own zip)
- zip member structure identical (files-only, top-level `<name>/` prefix)
- optional-field dropping (`description`/`author` absent) and string-array `capabilities` handling identical

Dependencies: `zip`, `jq`, `sha256sum`/`shasum` — all preinstalled on `ubuntu-latest`. The workflow step and paths filter are updated. The live `registry.json` is untouched; CI regenerates it on merge as usual.


JordanTheJet:
Closing unmerged: this tool runs only in CI on ubuntu-latest where python3 is preinstalled, so the dependency cost is zero — and the python version's tomllib parser is more robust to manifest evolution than the bash line-parser. The bash port remains on the `bash-build-registry` branch if the tool ever needs to run in a user-facing flow.

PR #6, opened 2026-07-10
feat: add channel plugins to registry

JordanTheJet:
## Summary
- add the channel plugin packages from JordanTheJet/zeroclaw-plugins to the upstream plugin registry repo
- retarget publish workflow, plugin READMEs, and registry entries to zeroclaw-labs/zeroclaw-plugins release assets
- teach the registry builder about the websocket_client permission used by the Discord plugin
- refresh registry.json so the seed index contains all 11 plugins

## Validation
- python3 -m py_compile tools/build-registry.py
- registry builder smoke test with Discord manifest and websocket_client permission
- cargo fmt --check for every plugins/* crate
- cargo clippy --locked --all-targets -- -D warnings for every plugins/* crate
- cargo test --locked for every plugins/* crate
- cargo build --locked --release --target wasm32-wasip2 for every plugins/* crate


JordanTheJet:
Superseded. This re-proposed the 8 poll channels (now merged via #8) plus discord + line, so it conflicts. The 8 are published; discord + line now land as host-gated source-only plugins in #9 (with the `registry = false` publish opt-out so they don't break the publish run or get mis-published). Closing in favor of #8 + #9.

PR #7, opened 2026-07-10
feat: channel plugin catalog — 8 published poll channels + discord/line (host-gated source)

JordanTheJet:
Brings the full channel-plugin set from the fork into the upstream registry. Fast-forward over `main` (which had only `redact-text`). Supersedes #5 and #6 (this consolidates + extends them).

## Published now (installable on a released host — `wasi:http` only)

`registry.json` advertises these 8 poll channels + the reference tool; each is a `wasm32-wasip2` component that self-polls inbound over `wasi:http` and sends over it (TLS host-side):

- **telegram, bluesky, mattermost, twitter, reddit, mochat, wechat, notion** (+ **redact-text** tool)

Each mirrors its built-in channel via `provides` (reads `[channels.<id>.*]`, native-wins), has a pure host-tested core, and re-vendors `wit/v0`.

## Source only (NOT in the registry — needs an unreleased host capability)

Landed as source so they're ready to publish the moment their host capability ships; deliberately **absent from `registry.json`** so a stock host never tries to install something it can't run:

- **discord** — needs the host **`ws-client`** WebSocket capability (zeroclaw#8863). Full Gateway protocol (identify/heartbeat/resume/dispatch) over host-owned WS; 37 host tests.
- **line** — needs the host **webhook-ingress** capability (zeroclaw#8862). Verifies `X-Line-Signature` HMAC itself, decodes events → inbound, pushes replies; 7 host tests.

## Notes

- `wit/v0` is re-vendored to the current host WIT (adds the `webhook-path`/`parse-webhook` exports + the gated `ws-client` import); the 8 published plugins carry capability-gated webhook stubs, so they instantiate unchanged.
- Publishing the two host-gated plugins is a follow-up once zeroclaw#8862 / #8863 land: add their entries to `registry.json` + a release build.

PR #5, opened 2026-07-08
feat(channels): add 8 channel plugins (telegram, bluesky, mattermost, twitter, reddit, mochat, wechat, notion) + re-vendor wit/v0

JordanTheJet:
## Summary

Adds a **Telegram channel plugin** to the registry, and re-vendors `wit/v0` from the ZeroClaw host so channel plugins build against the current ABI.

### `plugins/telegram`
A Telegram channel plugin as a sandboxed `wasm32-wasip2` WIT component:
- `getUpdates` long-poll (short `timeout=0` polls so it never stalls `send`) + `sendMessage`, over the host's `wasi:http` via the blocking [`waki`](https://crates.io/crates/waki) client in the sync channel exports (no executor needed; TLS is performed host-side).
- Pure Bot-API core in `src/telegram.rs` (Update→message mapping, `sendMessage` payload, allow-list, 4096-char chunking) with host `cargo test` coverage; `src/lib.rs` is the thin component shim.
- `capabilities = ["channel"]`, `permissions = ["http_client", "config_read"]`, `provides = "telegram"` (mirror-ready; inert on hosts without the `provides` feature, where it loads as a novel `telegram` channel).

### `wit/v0` re-vendor
The vendored contract had drifted from the host: `channel.wit` was missing `configure` and the `inbound` import, and `inbound.wit` was absent. **Channel plugins could not build a working `configure` against the stale copy.** Re-vendored unmodified from the host `wit/v0` (channel.wit, inbound.wit, plugin-info.wit), per `wit/VERSIONING.md`.

## Validation

- `cargo test` (pure core): 11/11 pass — Update mapping, forum-topic reply targets, allow-list, payload shape, chunking, config parsing.
- `cargo build --release --target wasm32-wasip2`: produces a valid component (360 KB).
- **Host instantiation**: the component instantiates under the live host (wasmtime 45 / `wasi:http@0.2.6`) with `http_client` granted — the `wasi:http` import resolves — verified via the host's `WasmChannel` (network-free: empty token skips `getMe`).

Text messages are supported; rich formatting, media, and inline-keyboard approvals are future work. Behavioral chat testing needs a live bot token; the runtime path is exercised by the host's channel-plugin E2E.

## Notes

- The channel-plugin runtime + `provides` mirror support land in the host via zeroclaw-labs/zeroclaw#8850 (channel runtime #8852, `provides` #8855). Until the channel runtime ships, an installed channel plugin is inert on the host.
- Author: @JordanTheJet

PR #9, opened 2026-07-10
feat(channels): add discord + line (host-gated, source-only) + registry opt-out

JordanTheJet:
## Summary

Brings the two remaining channel plugins from the fork into upstream — **discord**
and **line** — as **host-gated, source-only** components, and adds the publish
mechanism that makes that safe.

Upstream already publishes the 8 poll channels + `redact-text` (#8). These two are
different: each needs a host capability no released ZeroClaw has yet, so they land
as **source in the repo, absent from `registry.json`** — ready to publish the
moment their capability ships, but never offered to a stock host that can't run
them. (This is the plan from the closed #7, now done cleanly on top of the merged
`main`.)

## Why they can't just be dropped in

The publish pipeline builds and indexes **every** `plugins/*/`. Landing these two
as-is would:

- **break the publish run** — discord declares `permissions = ["websocket_client", …]`;
  `websocket_client` isn't in the host manifest schema yet, so `build-registry.py`
  rejects it and `sys.exit(1)`s the **whole** publish job (verified locally).
- **mis-publish line** — its manifest is valid, so it would be auto-added to
  `registry.json` and become installable on hosts that can't serve its webhook route.

## The mechanism — `registry = false`

A manifest opt-out for host-gated plugins:

- `build-registry.py` skips `registry = false` **before** validation, so a
  not-yet-known permission never fails the run; the plugin is never packaged or indexed.
- `publish.yml` skips building/staging it entirely.
- The host ignores unknown manifest fields (no `deny_unknown_fields` on
  `PluginManifest`), so the flag is inert host-side.
- README documents the convention.

Drop the line (or set `true`) to publish once the capability lands.

Verified locally: with the flag, discord+line are skipped and the run exits 0
(a normal plugin still indexes); without it, discord's permission hard-fails the run.

## `plugins/discord` — needs host `ws-client` (zeroclaw#8863)

Full Discord **Gateway** protocol (connect, IDENTIFY, heartbeats, RESUME,
`MESSAGE_CREATE` dispatch) over the host-mediated WebSocket; replies over the REST
API via `wasi:http` (TLS host-side). Pure protocol core in `src/discord.rs`
(frame parse, payload builders, mention gating, allow-lists, session state
machine, chunking) — **37 host `cargo test` cases**; `src/lib.rs` is the thin shim.
`provides = "discord"` mirrors the built-in channel. **Builds clean** against the
current `wit/v0` → `discord.wasm` (350 KB).

## `plugins/line` — needs host webhook-ingress (zeroclaw#8862)

LINE Messaging API channel: serves inbound over the host webhook route
(`/plugin/line`), verifying the `X-Line-Signature` HMAC-SHA256 itself (pure-Rust
`hmac`/`sha2`), and pushes replies over `wasi:http`. Pure core in `src/line.rs`
(config parse, signature verify, event→inbound decode, push-body build, chunking)
— **7 host `cargo test` cases**; `src/lib.rs` is the thin shim.
`provides = "line"`. **Builds clean** against the current `wit/v0` → `line.wasm` (324 KB).

## Validation

- `cargo test`: discord 37/37, line 7/7 (pure cores, no network, no wasm).
- `cargo build --target wasm32-wasip2 --release`: both produce valid components against upstream `wit/v0`.
- `build-registry.py` skip behavior verified both ways (see above).

## Scope

Only `plugins/discord`, `plugins/line`, and the publish mechanism + README —
no `registry.json` change, no changes to the 8 published plugins or `wit/v0`
(upstream's already carries the gated `ws-client` import and `webhook-ingress`).

This supersedes the discord/line portions of the now-conflicting #5 and #6, which
re-proposed the already-merged 8; those can be closed.


JordanTheJet:
Superseded by #10, which merged the Discord/LINE host-gated source work along with the broader Phase 4 channel plugin batch.

PR #54, opened 2026-07-18
feat: add audited Solana payment plugins — request, build, and confirm

Fianko-codes:
**Request → build → confirm.** A ZeroClaw agent can issue an invoice, construct a verifiable Solana token transfer, and **confirm the money actually arrived** — and no component in the path ever holds a private key, signs, or submits.

```
  solana_pay_request          spl_transfer_build              solana_pay_confirm
  ──────────────────          ──────────────────              ──────────────────
  invoice ──► solana: URL     ──► unsigned v0 tx ──► human ──► ──► reconciles the
              + derived           (all-zero sig       signs      recipient's own
              reference           slot, simulated)    off-band   balance delta
                    │                                                   ▲
                    └───────────── same derived reference ───────────────┘
                                   (not an argument — see below)
```

> ### A confirmer that accepts a `reference` argument is a confirmation-forgery primitive.
> A model that chooses the reference can point it at any payment on chain and get back `paid: true`. **This one derives it** — from the same four invoice fields, through the same shared-core function the request tool uses. There is no `reference` in the schema and unknown fields are refused, so the confirmation cannot be redirected. Ask with `amount: "1.4"` instead of `"1.5"` and the derived reference changes, so the scan looks at a different address entirely. The binding is structural, not a comparison a later refactor could skip.

And it confirms what **arrived**, not what was asked for: it reconciles the recipient ATA's pre/post balance delta against the invoice, so a hidden Token-2022 transfer fee cannot make an under-payment read as paid.

### Proven against live `mainnet-beta`, through the real ZeroClaw host

Not fixtures, not a mock host — ZeroClaw 0.8.3 with the Cranelift/Wasmtime plugin host, driven through `zeroclaw agent` against `api.mainnet-beta.solana.com`:

- **Real USDC transfer built and simulated on mainnet** — `err: null`, `unitsConsumed: 22853` — then re-decoded and **independently re-simulated** by `solders`, a different implementation from the plugin's own verifier.
- **Real mainnet PYUSD (Token-2022) refused, twice** — the second time with `allow_token_2022 = "true"`, because walking the **live** TLV list found `transferFeeConfig`, `permanentDelegate`, and `transferHook`. Policy earning its keep against real data.
- **A real finalized mainnet USDC payment decoded by the production path** ([`3yrMvnqX…BLGuik`](https://explorer.solana.com/tx/3yrMvnqXgMaukWqBi7heAn1ZqsoWmWhmivWwU1AbKhX7cRWCL5PBn3krUPvkpQrKoUL6dpUCbibUvX7CYqBLGuik), slot 436144302) — ATA re-derived locally and matched against the account actually credited, delta reconciled, then **refused for exactly one reason**: no invoice reference, because it wasn't made against one of our requests. The reference gate runs last, so reaching it proves token-program, destination, mint, decimals, and amount all passed on real mainnet bytes.
- **363-second delayed signing on devnet, then a finalized transfer** ([`2YrwNvTA…TEE8fL`](https://explorer.solana.com/tx/2YrwNvTAvM29ZsSXt9EVYmHsHizAYNCvusdkMr7prN9m1kKf21NNGrVBzfrW1mSJ8VLu1ouNyz3CxX26CxTEE8fL?cluster=devnet)) — the durable-nonce answer to blockhash expiry in an approval queue, with the executed message bytes **byte-identical** to what the plugin returned (`sha256 c70e5f40…`).

### Merge-readiness

| | |
|---|---|
| **CI** | **Green `Validate Required Gate`** — every job including Package dry run: [run 30541241327](https://github.com/Fianko-codes/zeroclaw-plugins/actions/runs/30541241327) |
| **`registry.json`** | Untouched — and `--check-history` is why that is the *only* correct answer, not caution. The packaging contract is exercised instead: reproducible zip digests, the exact entries `publish.yml` will emit, a real end-to-end registry install over HTTPS, and a fail-closed digest test. |
| **WIT** | Vendored `wit/v0` verified **byte-identical** to upstream `zeroclaw-labs/zeroclaw@e112ce6b`; `wit/` not modified. |
| **Build** | Clean-clone `--locked` `wasm32-wasip2` release build from an empty `CARGO_HOME`, all three plugins, pinned Rust 1.96.1. |
| **Tests** | **223** host-run, RPC mocked, no network, no wasm toolchain required. |
| **Custody** | **T0–T1 only.** No signing path, no submission path, no key material, at any tier. |
| **Audits** | Three independent adversarial audits — **0 Critical, 0 High**. Findings fixed with a regression test each. |

Reproduce the entire matrix in one command, with no wallet, network, or funds:

```bash
git clone --depth 1 https://github.com/Fianko-codes/zeroclaw-solana && cd zeroclaw-solana
./demo/reproduce.sh        # pinned toolchain, --locked, both clippy targets, wasm release builds → 14/14 PASS
```

---

## Components

| Component | Track | Custody | What it does | Permissions |
|---|---|---|---|---|
| **`plugins/solana-pay-request`** | A | **T1 · zero network, zero secrets** | Deterministic Solana Pay `solana:` transfer-request URLs. The reconciliation `reference` is a domain-separated SHA-256 over recipient, asset discriminator, canonical amount, and invoice id — so the same invoice always yields the same 32-byte key and a later reconciliation can find the payment. No RPC; no QR art in model context. | `config_read` |
| **`plugins/spl-transfer-build`** | A | **T1 · build & verify only** | One unsigned v0 SPL-token transfer: mint allowlist, per-mint caps, guarded idempotent ATA creation, `TransferChecked`, optional memo/reference — plus an **approval summary decoded from and re-checked against the exact bytes it returns**. Simulates; never signs, never submits. | `http_client`, `config_read` |
| **`spl-transfer-build` durable-nonce mode** | A | **T1 · operator-gated** | Optional transaction-lifetime mode selected only by operator config, never by the model. `AdvanceNonceAccount` at instruction 0, message blockhash == stored nonce, `authority == sender == fee payer` single signer. Survives a human approval queue that outlives the ~60–90 s blockhash window. | unchanged |
| **`plugins/solana-pay-confirm`** | A | **T0 · read-only** | Answers whether an invoice was actually paid. Takes the same four fields as the request tool and **re-derives the reference**. Decodes raw `base64` transaction bytes locally (legacy **or** v0), requires the reference on the transfer instruction itself as a read-only non-signer, and reconciles the recipient's own balance delta. Optional two-endpoint agreement; stateless, so the verdict is idempotent. | `http_client`, `config_read` |
| **[`nanosol`](https://crates.io/crates/nanosol)** (shared core, on crates.io) | E | — | Minimal `wasm32-wasip2`-native Solana core: base58/pubkeys, PDA + ATA derivation, exact decimal↔raw `u64` amounts (no floats in money paths), SPL Token / Token-2022 TLV mint policy, v0 message + transaction serialization, strict 80-byte nonce-account parsing, JSON-RPC shaping. **No `solana-sdk`, no `solana-client`.** | — |

**Recent blockhash remains the default.** Durable mode is additive and off unless an operator sets `blockhash_mode = "durable_nonce"` plus `nonce_account_pubkey`.

## Evidence index

Every claim above maps to a command that was run and an output that was observed. Failed commands are preserved in-tree rather than removed.

| Claim | Where |
|---|---|
| Recent-blockhash build, tests, devnet acceptance | [`spl-transfer-build/EVIDENCE.md`](plugins/spl-transfer-build/EVIDENCE.md) Part 1 |
| Durable nonce: 363 s hold, devnet acceptance, controlled failure | [`spl-transfer-build/EVIDENCE.md`](plugins/spl-transfer-build/EVIDENCE.md) Part 2 |
| Live mainnet read-only run, independent decode + re-simulation, Token-2022 refusals | [`spl-transfer-build/EVIDENCE.md`](plugins/spl-transfer-build/EVIDENCE.md) Part 3 |
| Packaging contract, reproducible digests, real registry install, fail-closed digest test | [`spl-transfer-build/EVIDENCE.md`](plugins/spl-transfer-build/EVIDENCE.md) Part 4 |
| Request-tool acceptance and real-host integration | [`solana-pay-request/EVIDENCE.md`](plugins/solana-pay-request/EVIDENCE.md) |
| Confirmer validation, mainnet host run, real-mainnet-bytes verification, **and what is not yet proven** | [`solana-pay-confirm/EVIDENCE.md`](plugins/solana-pay-confirm/EVIDENCE.md) |
| Threat models, config keys, worked examples, injection transcripts | each plugin's `README.md` |

<details>
<summary><b>Proof, in detail</b></summary>

### 1. 363-second delayed signing, then a finalized devnet transfer

The claim durable-nonce mode exists to support is "an unsigned transaction can sit in a human approval queue longer than a blockhash lives, and still execute." That was tested for real, not argued:

| Step | Observed |
|---|---|
| T0 — plugin builds unsigned durable tx | `2026-07-19T02:13:16Z`, 550 bytes, one all-zero signature slot |
| Held | **363 s** (a recent blockhash in the control run was already dead by 256 s) |
| Nonce re-read immediately before signing | **unchanged** — `AN1TfxqdcRnPoXUaD6xY9eb7puVwAUorLJBbVaUFjThS`, authority == sender |
| Signed externally | spike helper, key never inside the plugin; signed message bytes **byte-identical** to the plugin-returned message (`sha256 c70e5f40…`) |
| Submitted externally | [`2YrwNvTA…TEE8fL`](https://explorer.solana.com/tx/2YrwNvTAvM29ZsSXt9EVYmHsHizAYNCvusdkMr7prN9m1kKf21NNGrVBzfrW1mSJ8VLu1ouNyz3CxX26CxTEE8fL?cluster=devnet) → `confirmationStatus: finalized`, `err: null` |
| Balances moved | recipient **0 → 1.5**, sender **200 → 198.5** |
| `getTransaction` message bytes vs plugin output | **equal** (`sha256 c70e5f40…`) — what executed on chain is what the plugin returned |

**Nonce state, before and after** (account `8RxDmwBibTWTKvwFUxXsvDfzmwYdhUaNP1VzSZ8He8Ho`):

```
before / across the entire 363 s hold : AN1TfxqdcRnPoXUaD6xY9eb7puVwAUorLJBbVaUFjThS
after execution                       : EWmLUejtik2VUVYKe9UNqxvZ3mrcHTbba1BBewRZzEi
```

The nonce advanced **only** on execution. That is the whole safety property: the proposal cannot silently expire, and it cannot silently replay.

### 2. Nonce consumption on later-instruction failure — the case we chose to prove against ourselves

Durable nonces have an unforgiving edge: once nonce validation succeeds, a *later* instruction failure still advances the nonce and still charges the fee. Rather than only documenting that, it was reproduced on a **separate** disposable nonce account (`7rtVuvFRhiUCcVHQeiEsqgQAb7UgCEiHWvRdK1qNugjn`), built outside the plugin — the plugin simulates first and refuses to return a failing transaction:

- signature `62xgHDbVe4yd849njVkxdb84kTXxDbFMe6Xt1bZ1DZvfHQeZJ9Dn6PFzwv4Gp6oH2y8V9xf7hn5MXzPWsfcPoQ9K`
- status `confirmed`, `err: InstructionError[1, {Custom: 1}]` — the transfer failed
- nonce **consumed anyway**: `9ExdJdLpwALYq2WqBUdXMo8yB8r6hGVc3CocMV2Gfjud` → `ETRFV7AXtzJ3vMX51Ws8KTaPdnw6rFXcW7YAwoRdetgL`

This is the exact warning carried in the durable approval summary and the README. The documentation matches an observed on-chain outcome, not a reading of the docs.

### 3. Real ZeroClaw WASM runtime invocation

A disposable ZeroClaw 0.8.3 config with the Cranelift/Wasmtime plugin host; the component was discovered with capability `Tool` and permissions `HttpClient, ConfigRead`, driven through `zeroclaw agent` by a deterministic OpenAI-compatible oracle:

```
M4_AGENT_OK nonce_account=8RxDmwBibTWTKvwFUxXsvDfzmwYdhUaNP1VzSZ8He8Ho
            nonce=AN1TfxqdcRnPoXUaD6xY9eb7puVwAUorLJBbVaUFjThS
            transaction_sha256=3ce3f84a1af0c63c53a5c919632f0696970468dabe3f300593f2ec0730b56d52
```

The component made real devnet RPC calls (mint `getAccountInfo`, nonce `getAccountInfo`, durable `simulateTransaction`) and returned a 550-byte unsigned transaction. An **independent decoder** — not the plugin's own verifier — confirmed the exact shape: one all-zero signature slot, v0, exactly one required signer (sender == fee payer), message blockhash == the nonce, **0** address-lookup tables, **0** trailing bytes, and exactly four instructions: `AdvanceNonceAccount` (System, data `04 00 00 00`) at index **0**, ATA `CreateIdempotent` (data `01`), `TransferChecked` (disc `12`, five accounts including the reconciliation reference), and Memo v3. Component logs carried only bounded phase labels.

### 4. Live on `mainnet-beta`, read-only

Because there is no signing path, the read paths can be pointed at mainnet with zero risk — so they were. Same real ZeroClaw host, `rpc_url` set to `https://api.mainnet-beta.solana.com` (`solana-core 4.1.0`):

- **Legacy SPL (real USDC `EPjFWdd5…`)** — the plugin read the mint, read a blockhash, built a 485-byte unsigned transfer, and **simulated it on mainnet with `err: null`** (`unitsConsumed: 22853`). The bytes were then re-decoded by `solders` — an independent implementation, not the plugin's verifier — and **independently re-simulated against mainnet**: 1 all-zero signature slot, v0, 1 required signer, 0 ALTs, `CreateIdempotent` + `TransferChecked` (disc `12`, raw `1500000`, decimals `6`) + Memo, every account re-derived and matched.
- **Real mainnet Token-2022 (PYUSD `2b1kV6Dk…`)** — refused. Then refused *again* with `allow_token_2022 = "true"`, because walking the **live** TLV list found `transferFeeConfig`, `permanentDelegate`, `transferHook`, and confidential transfer — outside the supported safe subset. A transfer fee or permanent delegate is precisely how the amount a human approves diverges from the amount a recipient receives, so this is the policy earning its keep against real data rather than a fixture.
- **A free oracle fell out of it.** `nanosol` derives ATAs with hand-rolled sha256 + off-curve detection because `solana-sdk` will not build for wasip2. Mainnet checked that work: the derived canonical ATA **is** the account that actually holds the sampled wallet's USDC on chain, 4 for 4 across sampled wallets.

**Nothing was signed, nothing was submitted, and no private key for any mainnet address involved exists anywhere in this project.** Full record, exact addresses, and read-only reproduction: `plugins/spl-transfer-build/EVIDENCE.md` Part 3.

### 5. Closing the loop: a confirmation that cannot be pointed anywhere else

The reference is `SHA-256("zeroclaw-solana-pay-v1" ‖ recipient ‖ 0x01 ‖ mint ‖ u32be(len(amount)) ‖ amount ‖ u32be(len(invoice_id)) ‖ invoice_id)`, computed by the **same shared-core function** the request plugin uses. A frozen golden vector is asserted from *both* plugins' suites, so the pair cannot drift apart silently:

```
recipient FnHyam9w4NZoWR6mKN1CuGBritdsEWZQa4Z4oawLZGxa · mint EPjFWdd5…Dt1v (6 dp)
amount "1.5" · invoice "412"  →  reference 3FrMXf9ucXff2biCaz5ehKYr1yguHjWvwLcy8ALDVEnw
url solana:FnHy…ZGxa?amount=1.5&spl-token=EPjF…Dt1v&reference=3FrM…VEnw
```

**Independent tie to a real on-chain payment.** The devnet acceptance transaction (§1's sibling, explorer-linked in `EVIDENCE.md`) carries reference `GtaJ8kXf6UFmNKNkeYNhADCMFkxzShLCrsbvoJDkwK9J` in its `TransferChecked` account list. Deriving from that invoice's four fields reproduces that exact key — so the derivation this component performs matches a reference already attached to a settled payment on chain.

**The check that earns the tool its keep.** An earlier audit's one residual Medium was that a hidden Token-2022 transfer fee can make the amount *received* differ from the amount *transferred*. A confirmer that trusts the instruction amount inherits that hole. This one reconciles the recipient ATA's pre/post balance delta and requires it to equal the invoice exactly — `tests/adversarial.rs::a_token_2022_fee_shortfall_is_refused_even_though_the_instruction_amount_is_right` is that exact case: correct instruction amount, short delta, `paid: false`.

**Run against a real finalized mainnet payment.** `tests/fixtures/mainnet_usdc_payment.json` is a verbatim capture of public mainnet-beta responses for signature `3yrMvnqXgMaukWqBi7heAn1ZqsoWmWhmivWwU1AbKhX7cRWCL5PBn3krUPvkpQrKoUL6dpUCbibUvX7CYqBLGuik` (slot 436144302) — a v0 message with one SPL Token **`Transfer`** of 5 202 USDC base units, `pre 11491627208 → post 11491632410`. The production verification path decodes it, re-derives the recipient ATA locally and matches the account actually credited, and reconciles the delta. It then **refuses it, for exactly one reason**: no invoice reference is attached, because this payment was not made against a request from `solana-pay-request`. Since the reference gate runs *after* the token-program, destination, mint, decimals, and amount checks, reaching it proves all of those passed on real mainnet bytes.

That capture also settled a design question that fixtures could not: real wallets use the plain `Transfer` encoding, which names no mint and asserts no decimals. A decoder that only accepted `TransferChecked` would have refused real payments. The destination ATA was derived by an **independent Python implementation** of Solana's PDA derivation, including Ed25519 point decompression for the off-curve check — so ATA agreement is a cross-check, not self-consistency.

**Real host, real component, real mainnet.** One `zeroclaw agent` run against `api.mainnet-beta.solana.com`, driven by a deterministic oracle:

```
M5_AGENT_OK paid=false reference=2m1xGkYy7wzQahXH8rdCTS47AoGXovRGHwL7wA9iCHQ2
            expected_raw=1500000 match_count=0 bytes=645
```

`expected_raw = 1500000` for `amount = "1.5"` is only derivable from `decimals = 6`, and that came from the **live USDC mint account** — it is in neither the config nor the arguments. The oracle also asserted the schema the host advertised to the model: exactly `["amount", "invoice_id", "mint", "recipient"]`, no `reference`, no `__config`. The reference was reproduced byte-for-byte by an independent Python implementation.

**What is *not* proven, stated plainly:** a first-party live `paid: true`. Public devnet has pruned the earlier acceptance payment (`getTransaction → null`, first available block 479066362), the faucet rate-limited every airdrop attempt, and `solana-test-validator` will not run on the dev machine's pre-AVX2 CPU. `plugins/solana-pay-confirm/EVIDENCE.md` records all three blockers and the exact remaining command sequence. Nothing in this PR or the tree claims otherwise.

### 6. Adversarial verification of the headline claims

The load-bearing security claim is *summary equals bytes*. It is enforced by decoding the exact serialized transaction, semantically re-deriving every field, re-deriving both ATAs, and requiring a byte-equivalent canonical recompile of the whole message.

For the builder, an independent harness threw **11 byte-/structure-level mutations** at the production `verify_final_bytes` entry point — reordered instructions, appended duplicate, appended unknown program, recipient swap, amount divergence, summary/bytes divergence, mode confusion. **0 bypasses.** All durable final-byte mutations fail closed too.

For the confirmer, **24 adversarial fixtures** were run against the production entry points — each one a transaction a naive confirmer would accept, or an argument set an attacker would like accepted: a caller-supplied `reference`, arguments asserting `paid`/`signature`/`match_count`, a `__config` spoof, a recipient swapped off the allowlist, the reference present in the transaction but on another instruction, the reference attached as *writable*, a different mint into the expected ATA, ±1 base unit, wrong asserted decimals, a correct instruction amount with a short balance delta, an over-payment, a post balance with no increase, a balance record for the wrong index or a foreign owner, two transfer instructions, a failed transaction at both list and metadata level, a commitment below the operator minimum, an absent commitment, slot disagreement, address-table lookups, a second endpoint that disagrees or has never seen the signature, transport faults, endpoint error prose, and reference spam ahead of the real payment. **0 bypasses**, and each refusal carries the specific reason claimed rather than a generic one.

### 7. Tests, CI, and build parity

| Crate | Tests |
|---|---|
| `nanosol` | **60** (43 + 17 for the settled-payment read primitives) |
| `solana-pay-request` | **30** (29 + the cross-plugin golden reference vector) |
| `spl-transfer-build` | **71** (37 recent-mode, unchanged by durable work + 34 durable) |
| `solana-pay-confirm` | **62** |

All host-run, RPC mocked, no network, no wasm toolchain needed. Under the pinned CI toolchain, for all three crates: `cargo +1.96.1 fmt --check`, `test --locked`, `clippy --locked --all-targets -- -D warnings`, `clippy --locked --target wasm32-wasip2 -- -D warnings`, `build --locked --target wasm32-wasip2 --release` — **all PASS**.

- Green **`Validate Required Gate`** on the promoted tree — every job, including Package dry run: [run 30541241327](https://github.com/Fianko-codes/zeroclaw-plugins/actions/runs/30541241327) — Format, Registry contract, WIT drift, Plan matrix, Components shards 0–3, Package dry run, and the required gate itself, every job green.
- Strict `tools/ci/validate_components.sh` under a fresh `CARGO_HOME` and isolated target dir: `rc=0`, source-mutation guard clean.
- **Clean-clone check:** `git clone --depth 1` of this branch into an empty `CARGO_HOME` built for `wasm32-wasip2` `--locked`, resolving `nanosol 0.1.0` from crates.io. The pin is not a local-path illusion.
- Vendored `wit/v0` verified **byte-identical** to upstream `zeroclaw-labs/zeroclaw@e112ce6b`; `wit/` was not modified.

</details>

<details>
<summary><b>Security audit and residual risk</b></summary>

An independent read-only adversarial audit was run against the submission, a second focused audit against the durable-nonce diff, and a third against the confirmer and the core read delta. **Zero Critical, zero High** in all three.

The audit did not read for style — it built concrete attacks first and reported what survived. Confirmed properties: no signing / submission / private-key path anywhere; the single-signer invariant is *structurally* impossible to violate (a second required signer cannot be compiled into the message); durable `verify_final_bytes` rejects every divergence attack constructed against it; mode selection is operator-only (`deny_unknown_fields` + the host `__config` strip); a total, non-panicking strict parser; bounded output and error surfaces; recent-blockhash mode provably unchanged; and an approval summary that does not overclaim.

The third audit raised three findings against the confirmer, **all fixed in this head with a regression test each**:

- **Medium, documentation.** The README understated the RPC boundary. It said a dishonest endpoint could *hide* a payment; in fact the bytes, the reported commitment, and the balance metadata all come from the endpoint, so a single endpoint willing to fabricate all three consistently can forge a **positive** confirmation. That is now stated in a pull quote, together with why in-plugin Ed25519 verification would not close it — forged bytes never have to be real, so they can be signed with a key the attacker owns, and the delta remains the endpoint's word either way — and what does help: two-endpoint agreement, and the fact that every paid verdict names a signature an operator can check independently.
- **Low, API.** The public `verify_record` took the commitment level as an argument, leaving the gate in the caller. It now reads the level from the candidate and treats an absent status as unknown, so no caller can bypass it.
- **Low, availability.** A 25-signature window across two endpoints permitted fifty capped reads — up to ~12.5 MiB of parsing against a 1e9 fuel default with no wall-clock timeout, which would have surfaced as a plugin *fault* rather than a refusal. A single 1 MiB per-call read budget now bounds the whole call and refuses cleanly.

What remains is disclosed rather than closed, because it is not ours to close:

| Residual | Status |
|---|---|
| **A dishonest RPC endpoint can forge a positive confirmation** for the read-only confirmer, and can misreport mint or nonce state to the builder (Medium) | Documented prominently in all three READMEs, including why signature verification would not close it. Narrowed by optional two-endpoint agreement, and by the fact that every paid verdict names an independently checkable signature. For the builder it stays bounded because nothing is signed. Every transfer field is derived from operator policy, not from the model. |
| The plugin relies on the host stripping caller-supplied `__config` (Low) | Identical to the T1 boundary every config-reading plugin has; stated explicitly in the threat model. |
| Stateless host store → per-**call** caps are enforceable in-plugin, per-**day** caps are not (Low) | Stated honestly instead of faked. |

</details>

<details>
<summary><b>Bounty requirement compliance</b></summary>

| Requirement | Status | Where |
|---|---|---|
| One component, one tool | ✅ | `solana_pay_request`, `spl_transfer_build`, `solana_pay_confirm` — one tool each |
| Canonical plugin layout, matching `plugins/redact-text` | ✅ | standalone crate per plugin, own `[workspace]`, committed `Cargo.lock` |
| `crate-type = ["cdylib", "rlib"]` | ✅ | all three `Cargo.toml` |
| Pure core / thin wasm shim | ✅ | logic in wasm-free modules (`src/pay_request.rs`, `src/transfer.rs`, `src/confirm.rs`, `nanosol`); `src/lib.rs` is a `cfg(target_family = "wasm")` shim only |
| Host-run tests, providers mocked, no live network | ✅ | **223** tests total; RPC behind a mockable transport. The one committed real-network capture is a verbatim mainnet response replayed offline. |
| Clean `wasm32-wasip2` release build | ✅ | all three plugins, `--locked`, under pinned Rust 1.96.1 |
| Structured `log-record` logging, never stdout | ✅ | bounded phase/refusal labels only — no args, URLs, account or tx bytes |
| Minimal manifest permissions, used-only | ✅ | `solana-pay-request` = `config_read` **only** (imports no `wasi:http` at all); the other two = `http_client` + `config_read` for documented JSON-RPC methods |
| README: what it does, config keys, custody tier, threat model, worked example | ✅ | all three READMEs, incl. runnable worked examples |
| Prompt-injection test that fails closed + README transcript | ✅ | `tests/injection.rs`, `tests/component_and_injection.rs`, `tests/config_and_injection.rs`; transcripts are **generated by an asserted test**, so they are reproducible rather than a screenshot |
| Custody tier declared | ✅ | **T0** for `solana-pay-confirm`, T1 for the other two. Nothing above T1 anywhere; no component can sign or submit |
| License | ✅ | `license = "MIT OR Apache-2.0"` in every `Cargo.toml`, matching `redact-text` and `telegram`. Because this repository carries no root `LICENSE`, each plugin also ships `LICENSE-MIT` + `LICENSE-APACHE` — say the word and they come out. |
| `registry.json` untouched | ✅ | and for the right reason — `--check-history` structurally rejects contributor-added entries; the packaging contract and the install path are proven instead |

</details>

<details>
<summary><b>What fought us on <code>wasm32-wasip2</code></b></summary>

This is where most of the engineering actually went.

- **`solana-sdk` / `solana-client` do not build for a WASI Preview 2 component** (socket and `std::net` assumptions). So the primitives are hand-rolled in `nanosol` — but hand-rolled crypto and serialization is exactly the kind of confidence that deserves none. Every construction path is therefore compared **byte-for-byte against the official Solana crates used as host dev-dependency oracles**: `solana-message 4.3.0`, `solana-transaction 4.1.5`, `solana-pubkey`, `solana-nonce 3.2.0`, `solana-system-interface 3.2.0`, `solana-short-vec 3.2.2`, `spl-token-interface 3.0.0`, `spl-token-2022-interface 3.1.1`, `spl-associated-token-account-interface 2.0.0`. Golden vectors, not self-consistency.
- **Two `wit-bindgen` versions coexist in the graph**: 0.46 direct, 0.34 transitively via `waki 0.5.1`, which vendors its own and emits older `wasi:http@0.2.x` imports. Both link. This is real Cargo resolution — also present in the checked-out `telegram` reference plugin — and it was left visible rather than papered over with an override.
- **A component that imports `wasi:http` without the `http_client` grant will not instantiate.** That is why `solana-pay-request` has *no* HTTP path at all rather than an unused one: it genuinely imports nothing and declares `config_read` only.
- **The host creates a fresh store per call**, so tool plugins are stateless by construction. Per-call caps are enforceable in-plugin; per-day caps are not. Documented as a limitation instead of pretended away.
- **The host enforces no output-size cap and no wall-clock timeout**, so context-budget discipline is the plugin's job: output is shaped and **length-asserted in tests** so one tool call cannot blow up an agent's context.
- **`bincode` tolerates trailing bytes**, which makes it the wrong tool for parsing account state you do not control. The nonce-account reader is a strict fixed-80-byte parser that rejects Legacy version 0, `Uninitialized`, unknown version/state discriminants, and any other length.
- **`simulateTransaction`'s `sigVerify` and `replaceRecentBlockhash` are mutually exclusive**, and whether that even applies to a *durable* transaction is undocumented. Established empirically: durable simulation needs `sigVerify=false` **and `replaceRecentBlockhash=false`**, and the plugin additionally refuses if the RPC reports an unexpected `replacementBlockhash`. Stale nonce and wrong authority both surface identically as `BlockhashNotFound` — a fact that had to be discovered rather than read.
- **No floats anywhere in money paths.** Exact decimal-string ↔ raw `u64` conversion, tested at 0/2/6/9 decimals and at exact cap boundaries.
- **`curve25519-dalek` with `default-features = false, features = ["alloc"]`** for the off-curve check in ATA derivation — the only way that dependency fits a wasip2 component.
- **The existing transaction decoder could not be reused for settled payments.** `decode_unsigned_v0_transaction` requires a v0 message with all-zero signature slots — precisely what a landed transaction never has. Confirmation needed a decoder that accepts a *signed* legacy **or** v0 message, and one that accepts the plain `Transfer` encoding as well as `TransferChecked`. That second requirement was not a guess: the real mainnet capture uses `Transfer`, so a `TransferChecked`-only decoder would have refused real payments. Instructions with a transfer discriminant that fail to decode strictly are errors rather than skipped instructions, so a hostile transaction cannot hide a transfer behind a decoder that gives up.
- **Devnet history is not durable evidence.** The public devnet endpoint had pruned the acceptance transaction 12 days later, which is why the confirmer's live proof is a mainnet read-only run plus a committed mainnet capture rather than a replay of the earlier payment. Faucet rate limits and a pre-AVX2 CPU (`solana-test-validator` aborts on missing AVX2) closed the two obvious workarounds. Recorded as blockers, not glossed over.
- **Release artifact bytes vary with embedded absolute source paths** (703,484 bytes locally vs 706,776 from the validator's isolated rebuild). Recorded hashes therefore identify a specific clean-clone build environment; they are honestly *not* presented as a universal reproducible-build hash.
- **CI pins Rust 1.96.1 while the dev box runs 1.97**, so the entire matrix is run under `+1.96.1`. Related sharp edge: a devnet faucet tool's own locked `wasm-bindgen 0.2.86` rejects 1.96.1 — installed unlocked and disposable, never added to either repository.
- **The host itself can persist tool output even when component logs do not.** Found during a `--verbose` run. Because the unsigned transaction is intentionally load-bearing tool output, the README tells operators to set `observability.log_tool_io = "off"`.
- **Plugin discovery alone is not enough** for a custom model endpoint — `native_tools=true` is required before the tool schema reaches the model. Worth knowing before anyone tries to reproduce the host run.

</details>

<details>
<summary><b>Packaging: <code>registry.json</code> is untouched, and here is the proof it doesn't need to be</b></summary>

`registry.json` is a generated index, and I did not hand-edit it. That is not caution — it is the only correct answer:

```
$ python3 tools/build-registry.py --check-history registry.json dist/registry.json
error: registry entry was added outside the publication builder: solana-pay-request@0.1.0
error: registry entry was added outside the publication builder: spl-transfer-build@0.1.0
```

`--check-history` is part of the **required gate**, and CI runs it with the merge base's `registry.json` against the committed one. Any entry a contributor adds fails the build. Only `publish.yml`, on push to `main`, may add entries — and it does so automatically, uploading the zips, verifying every URL and digest, and committing the refreshed index itself.

So instead of a hand-edit, the packaging contract was **exercised**:

- **The CI package dry run was reproduced locally.** Reproducible zips (fixed 1980-01-01 timestamps and permissions): `solana-pay-request-0.1.0.zip` `sha256 f111d919…`, `spl-transfer-build-0.1.0.zip` `sha256 d577ad1c…`. Packaging the same staged bytes with two **different** `--release-base` values produced **identical digests** — only the `url` differs, so the index churns only when plugin content actually changes. `--check-metadata` and `--check-publication` both pass; `plan_matrix.py` selects exactly these plugins.
- **The exact entries your publish workflow will emit are recorded** in `plugins/spl-transfer-build/EVIDENCE.md` Part 4, so merge requires no follow-up from you.
- **A real end-to-end registry install was demonstrated** over HTTPS from real release assets: `zeroclaw plugin search solana` finds them, `zeroclaw plugin install <name> --registry <url>` installs them, a pinned `@0.1.0` install works, and the host seeds the `[[plugins.entries]]` stanza.
- **Digest verification was proven to fail closed.** Against an index whose `sha256` was replaced with 64 zeros: `Error: plugin archive sha256 mismatch`, exit 1, `No plugins installed`.
- **One real discoverability bug was found and fixed.** `zeroclaw plugin search solana` did not return the transfer plugin, because its manifest description never said "Solana". It now reads *"Build and verify an unsigned Solana SPL token transfer with a recent blockhash or durable nonce"*. Description is manifest-owned metadata the publication builder may refresh; the immutable `name@version` identity and release fields are untouched.
- **A packaging hazard worth your attention:** re-cutting a release under the same tag and filename made GitHub's CDN serve the *deleted* release's bytes — the install succeeded with a stale manifest and `curl | sha256sum` returned the old digest. Your publish workflow already defends against this by refusing to overwrite an asset whose bytes differ; this is confirmation that the defense is load-bearing rather than theoretical. Filed separately as [#148](https://github.com/zeroclaw-labs/zeroclaw-plugins/issues/148) so it doesn't live only in this PR.

</details>

<details>
<summary><b><code>nanosol</code> packaging, revisions, and milestone record</b></summary>

### Packaging boundary — decided, not delegated

All three plugins depend on **[`nanosol 0.1.0`](https://crates.io/crates/nanosol), published on crates.io** — a normal registry dependency, not a Git revision of a personal repository and not a vendored copy duplicated into each plugin directory. Each plugin directory stays standalone and zip-installable, and the CI structure guard (which forbids a shared crate inside `plugins/`) is respected without a new top-level directory.

Verified by a `git clone --depth 1` of this branch into an **empty `CARGO_HOME`**, building `--locked` for `wasm32-wasip2` from scratch: the dependency resolves from the registry with nothing local in play. The flip from the earlier Git-revision pin moved **exactly one line per lockfile** — the `nanosol` source plus its checksum — and no other dependency version, which the byte-oracle tests would have caught if it had, since they fail loudly if the core's output shifts by a single byte.

If you would rather vendor the core into each plugin or carry it as an in-repo crate outside `plugins/`, both are a one-line dependency change plus a regenerated `Cargo.lock` — no plugin logic, no manifest, no WIT, no test changes.

Independently, the durable-nonce mode can be reverted wholesale (`git revert -m 1 3a6b263`), returning the branch to the frozen, separately-audited recent-blockhash-only submission at tag `m35-security-freeze-95e10dc`. Nothing here is a one-way door.

### Heads

| Repo | Ref | Revision |
|---|---|---|
| `zeroclaw-plugins` (this PR) | `agent/m25-solana-pay-request` | confirmer promotion merge `f1a0892`; durable-nonce promotion merge `3a6b263`; mainnet + packaging merge `d42aeb4` |
| `nanosol` core | [`nanosol 0.1.0`](https://crates.io/crates/nanosol) on crates.io | built from `Fianko-codes/zeroclaw-solana@2093879cd1cc28c6182386706d70b8a56b3b07be` |
| vendored WIT source of truth | `zeroclaw-labs/zeroclaw` | `e112ce6b5ccdac9e1cb166bab217e730dd7e24c2` |

### Immutable tags

- `m35-security-freeze-95e10dc` → `95e10dc1b8ec4c796b22d50ffc63136e462eaf0a` — audited recent-blockhash-only submission; the clean fallback.
- `m3-known-good-3f7f8e9` → `3f7f8e9a5db1a7d7c626d1f22ace166cd0d02b17` — pre-remediation known-good.
- `m4-promotion-candidate-ac7c749` → durable-nonce work as validated before promotion.

`3a6b263` is a normal merge commit (parents `95e10dc` + `ac7c749`), so `git revert -m 1 3a6b263` returns the branch to the frozen submission exactly.

### Reproduction and spikes

- **One command:** `demo/reproduce.sh` in the `nanosol` repository runs the exact CI command set — pinned toolchain, `--locked`, both Clippy targets, wasm release builds — across the core and all three plugins and prints a pass/fail table. 14/14 PASS with no wallet, network, or funds. `demo/qr.sh` renders a request URL as a terminal QR, printing the decoded terms first, which is where QR rendering belongs: outside the model's context window.
- Durable-nonce Phase-A spike results, PDA/ATA oracle spike, and the wasip2 build spike live in the `nanosol` repository.
- **Phase-A devnet signature** (SOL-transfer viability, prior to the SPL acceptance run): `3fdropMCZm1Yriy1iTupQTQpMcLxrciAC7YtaVGw7AC8k1avFgkjZUX4NK7wDsboCpSRPtNCFPxa5ps2WKoHK1ig`, nonce `4vMZqWuEMy9gAa5PWaYVWkQhvLKFquKH62fnzfKcvDkN` → `9ExdJdLpwALYq2WqBUdXMo8yB8r6hGVc3CocMV2Gfjud`.

### Private-key handling during validation

No private key ever entered `nanosol`, any plugin, any config, or any committed evidence. Disposable devnet keypairs lived only in a session scratchpad at mode `0600` and were destroyed. External signing used a spike helper outside the plugin. The plugin received only public keys, via operator config.

</details>

## Deliberately not in scope

- **No signing, no submission, no key custody, at any tier.** The agent proposes; a human disposes, through ZeroClaw's native `Supervised` approval gate.
- **Not built:** native SOL transfers, swaps, NFTs, multisig proposals, address-lookup tables, multi-recipient transfers, arbitrary instructions.
- **The confirmer is not a watcher.** It is stateless with no cursor, so duplicate-notification suppression and termination remain the SOP's business. It exposes `match_count` so an SOP can detect a **double payment**, which is a real merchant condition a cursor-based watcher silently skips. Confirmation covers SPL Token and extension-free Token-2022 only, one transfer instruction per transaction, top-level instructions only, and a bounded scan window.
- **Durable-nonce mode supports only `nonce authority == sender`.** A separate nonce-authority signer is intentionally unsupported — it would require a second signer, which the single-signer invariant is designed to make structurally impossible.
- **Token-2022:** extension-free mints only, even with operator opt-in. Any TLV extension or unknown discriminant fails closed. Discriminants are tested against the authoritative interface source (`Pausable = 26`, current extensions through 28), not a remembered list.
- Durable nonces are functional today, but Solana's docs carry a forward-looking "may be deprecated" notice (SIMD discussion #415 — a discussion, not an activated change) and the recent-blockhashes sysvar is deprecated for on-chain reads while still required as an `AdvanceNonceAccount` account. Operators adopting durable mode should track this.

---

*Submission for the Superteam Brasil "Solana-native plugins for ZeroClaw" bounty. `registry.json` is a generated index and was not hand-edited; nothing has been published. This is not a claim of upstream acceptance or maintainer endorsement — the packaging boundary above, and merge itself, are maintainer decisions. Review comments welcome on any part of it.*


Fianko-codes:
The description is now the single source of truth — rewritten around a short summary
with the detail collapsed underneath, and the earlier progress comments cleared so
there is only one thing to read.

No evidence was lost in the process: it moved into one `EVIDENCE.md` per plugin, so
each directory now matches the `redact-text` layout more closely instead of carrying
one to four milestone-named files.

Out of draft, ready for review, and the branch is stable — I'll only touch it in
response to review comments.

PR #25, opened 2026-07-17
feat: add policy-constrained Solana Pay request tool

Absamad-dew:
## Summary

- add the `solana_pay_request` WIT tool for policy-constrained native-SOL transfer-request URLs
- enforce operator-owned recipient allowlists and an exact checked-decimal SOL cap
- keep signing, broadcasting, RPC, balance reads, and settlement verification outside the component
- restack this PR on #24 so manifest metadata and immutable release validation use the finalized plugin stack

## Validation Evidence

Commands run on current head `ae8aebb`:

- `cargo fmt --manifest-path plugins/solana-pay-request/Cargo.toml -- --check`
- `cargo test --locked --manifest-path plugins/solana-pay-request/Cargo.toml` - 14 tests passed (1 identity + 13 behavior tests)
- `cargo clippy --locked --manifest-path plugins/solana-pay-request/Cargo.toml --all-targets -- -D warnings`
- `cargo build --locked --manifest-path plugins/solana-pay-request/Cargo.toml --target wasm32-wasip2 --release`
- `python3 -m unittest discover -s tools/tests -p 'test_*.py'` - 4 tests passed
- `python3 tools/build-registry.py --source-plugins plugins --check-metadata registry.json`
- immutable package simulation for `solana-pay-request@0.1.0` passed
- `git diff --check`
- source-built ZeroClaw 0.8.3 E2E with Ollama `qwen3:4b-instruct`: one native tool call, the current-head WASM component executed in 0.822 s, and the agent returned the expected Solana Pay URL

## Security & Privacy Impact

- New permissions, capabilities, or filesystem scope? `Yes`: one `tool` capability with only `config_read`.
- New external network calls? `No`.
- Secrets, tokens, or credentials handling changed? `No`.
- PII, real identities, or personal data in diff/tests/fixtures/docs? `No`.
- Publisher identity: manifest uses the public GitHub handle `Absamad-dew`; no personal name is embedded.
- The component has no key material and no signing/broadcast/RPC path. The generated URL asks a compatible wallet to compose a SOL transfer, so wallet review and approval remain required.

## Compatibility

- Backward compatible? `Yes`: this is a new unpublished plugin identity.
- Config/env/CLI surface changed? `Yes`: new `plugins.entries` keys for this plugin only.
- `solana-pay-request` is the manifest/package ID; `solana_pay_request` is the exported WIT tool name. Tests and README pin the distinction.
[...]

PR #63, opened 2026-07-18
feat(plugins): add Solana portfolio sentinel tool plugins

ZiBibro:
## Summary

Three self-contained Solana tool plugins that turn a ZeroClaw agent into a daily portfolio sentinel. Two read-only tools cover the watching side: `lending-health` reports how close DeFi borrow positions sit to liquidation, and `stake-monitor` reports delegation state, stake amounts, validator health, and epoch rewards for the operator's own stake accounts. The third, `stake-tx-build`, closes the loop from alert to action: it builds an unsigned delegate or deactivate transaction for a human to approve, and holds no key material of any kind.

Each plugin is a standalone `wasm32-wasip2` component in the `plugins/redact-text` reference format: a pure host-testable core, a thin `#[cfg(target_family = "wasm")]` shim, host-run tests over captured fixtures, and a README with a threat model and a live prompt-injection transcript.

**On the status of this PR.** Judging is over, so this is out of draft and ready for review. The branch was opened on 18 July, before the 22 July listing update that asked contributors not to open registry PRs during the bounty; it sat as a draft from then until judging closed. The code lives and is published at [ZiBibro/solana-portfolio-sentinel](https://github.com/ZiBibro/solana-portfolio-sentinel), and the comments below record what changed since: the typed-config migration, verification against a host built from current upstream master, and the defects that verification found. Note that CI has never run on this branch: every workflow run sits at `action_required`, waiting on a maintainer to approve a first-time contributor's run.

## Plugins

**lending-health** reads Kamino positions from the public Kamino REST API and decodes MarginFi positions straight from on-chain account state via `getProgramAccounts`, reading the program's own maintenance-weighted health cache at fixed byte offsets. Positions are classified against operator-set thresholds on the liquidation buffer, meaning the share of the distance to a position's own liquidation line that is still left. Kamino's documentation describes that gap as the quantity that matters, since the opening limit is already spent once a position exists. A position whose basis is missing, zero or non-finite is reported as UNKNOWN, and a position with no debt is reported as such, so no verdict is issued without a measurement behind it. Each line names the obligation it came from, and the report is capped so a recurring briefing never floods the agent context. Rows carry their basis in the label, because MarginFi's maintenance-weighted ratio and Kamino's plain ratio are not the same measurement. Drift is out of scope on purpose: its public API exposes no current health figure, and a reconstructed number would risk being wrong about someone's liquidation distance.

**stake-monitor** reports, per allowlisted stake account: delegation lifecycle status, stake amount, validator delinquency (fetched with a server-side `votePubkey` filter rather than the full roster), vote lag in slots against an operator-set threshold, epoch progress, and the previous epoch's reward. Commission comes from `inflationRewardsCommissionBps`, since the legacy `commission` field can be null. Vote lag is the early-warning half: a validator drifting behind the head is visible before it is formally delinquent. The header totals only stake still committed to a validator, because a cooled-down account keeps its delegation record on chain. When a reading is unavailable the row says so, rather than asserting a fact about the chain that was never established.

**stake-tx-build** encodes legacy messages by hand (header bytes, compact-u16 lengths, bincode discriminants, account metas) with no `solana-sdk` dependency. The instruction bytes and the per-position account order are locked by a golden test against a live mainnet delegate transaction ([5yaZiJMV…](https://solscan.io/tx/5yaZiJMVnN5fM5K4rHQFrntaprKQJJbuLqiVGWh7Dkg1MqtswUno83BTozmzN8xAfLZTtFTZiwhTUZsmNoa5kVRA)). Before building anything it verifies the endpoint's genesis hash against the operator's pinned cluster and refuses on mismatch. With an optional durable-nonce pair in config the transaction starts with `AdvanceNonceAccount` and survives an approval queue; without one, the summary states the roughly 60 to 90 second blockhash window. The pre-signing summary names every address in full and derives the signer count from the message that was actually serialized, so a nonce authority held on a separate key is disclosed as a second required signature.

Legacy rather than versioned is deliberate. Address lookup tables are what v0 buys, and these transactions carry four and seven accounts, so there is nothing to compress. The transaction ends in the operator's own wallet, where support for versioned transactions is still uneven.

## Custody and safety

Custody tiers on the ZeroClaw ladder: `lending-health` and `stake-monitor` are **T0 Read**, `stake-tx-build` is **T1 Build**. Nothing here signs or submits, and no private key ever appears in config or code. The most sensitive value any of them holds is an RPC endpoint URL.

Safety is structural rather than prompt-based. Every address a tool acts on resolves only against an operator-owned config allowlist, so a hijacked model can narrow a query but cannot widen it to an attacker's address; delegation stays disabled until the operator opts in with `allowed_vote_accounts`. Config parsing is fail-closed. Since 0.2.0 each manifest declares a closed `config_schema`, so an unknown or misspelled key is refused by the host before the component starts, and the keys a plugin cannot work without are `required` rather than defaulted: a withheld grant leaves a reader with no allowlist and it refuses to run rather than widening. A value carrying invisible characters is refused by codepoint and byte position, since a label that renders identically to an allowed one turns a refusal into nonsense.

Text a third party controls is treated as untrusted input to the model. Kamino product tags are length-bounded and narrowed to a small character class. The `message` field of a JSON-RPC error is relayed as an explicit quotation, stripped of control characters, with any quotation mark inside it folded so the wrapper cannot be closed early, and capped at 160 characters. The diagnostic survives and the foothold does not. Tool arguments reject unknown fields, and both the report and the failure paths are bounded by the same character cap. Each README carries a live fail-closed transcript of an injection attempt.

## Validation

Toolchain `cargo 1.97.1` / `rustc 1.97.1`, target `wasm32-wasip2`. Everything below was measured on 30 August 2026 at `38d4b24`, the head of this branch:

| plugin | host tests | wasm bytes | sha256 |
|---|---|---|---|
| lending-health | 94 | 400264 | `f075c6b8bedff760376714b50d50ccb11ce3a39200996f213b09942ea50b3330` |
| stake-monitor | 56 | 354078 | `32b4f4372452e746f04db9dce268788be4c6f2d5f1ea34a7510e3cc4d08bbbbe` |
| stake-tx-build | 91 | 380633 | `f685ce49a09a5c56d767d6dff29f92b3bebc6549365fc9fb6b204e97407867e6` |

One caveat on that table against the previous revision of it: the host tests were run on this branch, and the three components were built over a warm `target/`, not after `cargo clean`. Byte sizes are unaffected by that. The digests are a fingerprint of these particular builds, and the paragraph below still applies to them.

- `cargo fmt --all -- --check` clean
- `cargo test --locked` green: 241 host tests total, no network in tests
- `cargo clippy --locked --all-targets -- -D warnings` clean on the host target and on `wasm32-wasip2`
- `cargo build --locked --target wasm32-wasip2 --release` produces the component

**On the digests, precisely.** Two clean builds in different directories on this machine, one of them the mirror repository's own tree, reproduce each digest above byte for byte, and a third repeat of the first one matched again. That is the claim, and it is the whole claim. A build over a warm `target/` does not reproduce them: it can return an artifact of identical length with a different digest, so only clean builds are worth comparing. We have also seen a digest move across sessions on unchanged source at identical length while remaining stable within a session, and we did not isolate the cause, so we do not claim these digests are stable across machines or toolchain installs. Treat them as a fingerprint of this build rather than a universal constant, and re-measure locally before concluding anything from a mismatch. The reproducibility that does hold, and that matters more, is the source-to-behavior one: a clean clone builds, tests green, and produces working components.
- `tools/build-registry.py --source-plugins plugins --check-metadata registry.json` reports the three as valid pending unpublished sources; `registry.json` itself is untouched

Fixtures for `lending-health` and `stake-tx-build` are captures from live endpoints, including a mainnet transaction and a MarginFi account. One MarginFi fixture is synthetic and labelled as such in its doc comment, because no live capture carried a written maintenance pair. `stake-monitor` tests run against literals that mirror live replies.

Beyond the gates, the components were exercised in a source-built v0.8.3 host (`--features plugins-wasm,plugins-wasm-cranelift`) against live mainnet and devnet data. On devnet, `simulateTransaction` accepted both built transactions with `err: null`, at 10882 compute units for the deactivate and 16956 for the delegate. Four of the five delegation lifecycle states were observed on real accounts. The durable-nonce path was driven against a real initialized nonce account, together with its three failure paths.

Workflow runs from forks wait for maintainer approval in this repository, so no check will go green here on its own. The mirror repository runs the same gates on every push and its CI is green.

**On the relationship between the two trees, precisely.** As of 30 August 2026 the three plugin directories on this branch are byte-identical to the ones in the standalone repository, which is where the work has been done. That was not true while the branch sat parked, and the numbers in the table above were re-measured after the sync rather than carried over.

**What 0.2.0 changes, and why now.** [#147](https://github.com/zeroclaw-labs/zeroclaw-plugins/issues/147) makes `config_read` and `config_schema` a biconditional, and [zeroclaw-labs/zeroclaw#9126](https://github.com/zeroclaw-labs/zeroclaw/pull/9126) landed the enforcement on 20 August with no shim: a manifest that requests the permission without a schema is no longer discovered or installed. These three packages were on the wrong side of that, so this revision migrates them following the `redact-text` template in [#149](https://github.com/zeroclaw-labs/zeroclaw-plugins/pull/149).

It is breaking for operators, hence 0.2.0 rather than a patch: `wallets`, `protocols`, `stake_accounts` and `allowed_vote_accounts` were comma-separated strings and are JSON arrays now, and the numeric keys are real numbers. The guests read `__config` as typed JSON in one `serde` step and do no string parsing.

Two relations stay in the guest code, because JSON Schema cannot state them between sibling properties: `warn_liquidation_buffer` must exceed `critical_liquidation_buffer`, and the durable nonce pair must be set together or not at all. Numeric bounds are duplicated rather than delegated to the schema, since a host-side `cargo test` runs with no validation at all and a NaN satisfies every numeric keyword while failing the comparison a report is classified on.

Replacing the guest-side unknown-key check, which `additionalProperties = false` now does earlier and better, are two tests per plugin that read `manifest.toml` as text: one asserts `config_read` and `config_schema` appear together with a closed schema, the other asserts every key the guest reads is declared in it. Both were mutation-tested, against a renamed schema property and against a deleted `config_schema` table, and each failure names the offending key.

Config errors also stopped quoting the value that caused them. `serde_json::Error`'s `Display` embeds it, and every value here is a pubkey, an allowlist entry or the operator's own RPC endpoint; in `stake-tx-build` the `authority` names the account a built transaction would be signed by, so returning it to the model inside an error was the worst case available.

## Relationship to open PRs

Several submissions now touch adjacent ground, so a short map may save review time.

- `plugins/lending-health` is claimed by more than one PR. #84 decodes Kamino obligations on-chain at validated offsets, which needs a `getProgramAccounts`-capable endpoint. #78 and #80 each bundle a lending-health of their own beside other plugins. This variant covers Kamino together with MarginFi decoded from on-chain state, works against a plain public RPC, and refuses to run without a configured wallet allowlist.
- #104 sits at its own path, `plugins/liquidation-guard`, and goes deeper on Kamino than this suite does. It does not cover staking.
- `plugins/stake-monitor` is also proposed by #60 and #141.
- `plugins/stake-tx-build` has no counterpart among the open PRs.

Happy to rename directories, rebase onto whichever lending or stake variant maintainers prefer, or split this into separate PRs if that is easier to land.

## Design notes

- Standalone crates, no shared core. A common crate would either mint a new top-level convention (`libs/` or `crates/`) or couple the plugins; duplicating a small RPC helper per crate keeps every plugin independently reviewable and installable, which seemed the better trade for this repository today.
- Transaction assembly is hand-rolled on `bs58` and `base64`, with the byte layout pinned by a golden test against a real mainnet transaction. `waki` for blocking `wasi:http` and `serde_json` are the only heavy dependencies. The pure-core split is what made this testable: every decode, encode, and threshold decision lives in a plain Rust module the host tests exercise without a wasm toolchain, and the shim holds nothing beyond the host wiring.
- `stake-tx-build` re-checks a delegation target's live standing before it builds. On the delegate path it calls `getVoteAccounts` filtered by `votePubkey` and parses the reply into a `VoterStanding`, so a validator the chain currently lists as delinquent is named as such in the pre-signing summary. The allowlist stays the enforcement point, and this is the dynamic half beside it: an allowlist is static while delinquency is not. The RPC's own default hides delinquents that hold no active stake, which is why the request sets the flag that includes them.
- A note for operators behind TLS-intercepting antivirus or proxies: `wasmtime-wasi-http` trusts the bundled webpki roots only, so intercepted HTTPS fails with `TlsProtocolError` for any HTTP-using plugin. Documented here after hitting it during live testing; worth knowing before debugging a plugin that works everywhere except one machine. It is filed upstream as zeroclaw-labs/zeroclaw#9653, accepted the same day, with the maintainer asking that it be coordinated with the in-flight egress-policy work rather than landing a competing hook. Nine issues came out of this build. All nine were accepted upstream, five of them at P1, and three are already fixed: #9465 on the silent precheck boundary, closed as completed by merged PR #9478; #9642 on an approval timeout recorded as an operator denial, closed by PR #9423; and #9652 on `config set` refusing a cron alias that `config list` prints, closed as completed on 5 August by merged PR #9705. Six remain open: #9643 on the WIT versioning document, #9672 on `cron add` examples in the host's own help that do not execute as printed, and #9653, #9654, #9655, #9656 filed on 2 August, accepted that day with reproduction quality rated complete.

## Where this can go next

Extension points the design leaves room for, offered as the shape of a second
version rather than as a roadmap. Nothing here is planned work while judging runs.

- A `nonce-manage` companion. Creating a nonce account is a one-off setup step, and doing it with the Solana CLI is the ordinary way to do it. What such a companion adds is the rest of the lifecycle an approval queue eventually needs: rotating an authority, retiring an account, and keeping one nonce per pending transaction, since a single account serializes to a single in-flight transaction.
- Position history. Every report here is a snapshot by design: each run reads only what the chain says at that moment and holds no state between runs, which is what keeps a run reproducible and the component free of storage permissions. Trend is a real second question, and the shape that answers it is a companion that owns a datastore and reads these reports.
- MarginFi and Kamino share almost nothing structurally; a third protocol would be the moment to extract a common position model rather than inventing one prematurely.
- Refusing on delinquency rather than warning about it, if maintainers want the builder to enforce what the official CLI enforces. The Solana CLI errors by default when the vote account's root slot has fallen behind, unless `--force` is passed or the account carries no active stake; the builder today names the condition in the pre-signing summary and still builds.


ZiBibro:
**Update: cluster gate, vote lag, and honest health reporting (aae9aee).**

Since the branch was opened, three changes went in, each because a specific claim could be attacked:

- `stake-tx-build` now verifies the endpoint's genesis hash against the operator's pinned cluster before it builds anything, and refuses on mismatch. The README states plainly what that buys (an honest endpoint on the wrong chain, a config typo) and what it does not (a proxy that echoes the expected hash, a fork that inherits mainnet's genesis).
- `lending-health` no longer states a liquidation distance when MarginFi's maintenance pair is absent, and it only treats a cleared `HEALTHY` bit as a verdict when `ENGINE_STATUS_OK` shows the engine actually wrote the cache. A condemned position keeps CRITICAL and leads the report; a never-written cache condemns nothing. Each line also names the obligation it came from.
- `stake-monitor` reports validator vote lag against an operator-set threshold plus epoch progress, so drift is visible before formal delinquency, and it degrades those two readings instead of failing the whole report when the epoch reply is unusable.

Both plugins now bound the delivered payload and the error paths inside the same documented character cap, so a pile of server-controlled RPC error text cannot flood the agent context.

**Local validation.** Workflow runs from forks wait for approval here, so the gates are run locally after `cargo clean` for each package on `cargo 1.97.1` / `rustc 1.97.1`:

```
lending-health   fmt ok  clippy ok (host + wasm32-wasip2, -D warnings)  test 59 passed  wasm 385239 B
stake-monitor    fmt ok  clippy ok (host + wasm32-wasip2, -D warnings)  test 35 passed  wasm 348843 B
stake-tx-build   fmt ok  clippy ok (host + wasm32-wasip2, -D warnings)  test 36 passed  wasm 360284 B
```

sha256 of the release components, reproducible from a clean checkout:

```
caf22bc64e1c6d3f5c96a1c43360daefc0005a91c18e6042e9823f6e26702c3e  lending_health.wasm
71c5000451178628bbd471fc51a381f95d00ec973bd52efc64039b2b2a4aef85  stake_monitor.wasm
fb067b370cf264695d87f677b4c13294396fd447ac239b714509e8b30e7c0836  stake_tx_build.wasm
```

`tools/build-registry.py --source-plugins plugins --check-metadata registry.json` reports the three as valid pending unpublished sources, with `registry.json` untouched.

**On overlap.** Several submissions now propose plugins at `plugins/lending-health` and `plugins/stake-monitor`. I have no attachment to the directory names: happy to rename, to rebase onto whichever variant you prefer as the base, or to split this into separate PRs if that lands more easily. `plugins/stake-tx-build` has no counterpart among the open PRs, and it only makes sense next to a stake reader, which is why the three arrived together.

Also removed the repository-local custody-tier labels from the PR description after seeing the note on #25; the mechanics are stated directly instead.

ZiBibro:
The bounty listing was updated on 22 July with guidance I had not seen: registry PRs should not be opened during the bounty, and registry merges happen separately after judging by maintainer invitation. This PR predates that update (opened 19 July), so I am moving it to draft to keep it out of your review queue for the duration rather than closing it and losing the history.

The code lives in its own repository with CI running the same gates as this branch: https://github.com/ZiBibro/solana-portfolio-sentinel

Happy to mark it ready again after judging if this plugin family gets an invitation.

ZiBibro:
Judging is over, so this comes out of draft. Two things changed since it went in.

**Migrated to typed instance config.** All three manifests now declare a closed Draft 2020-12 `config_schema` and the guests consume typed `__config`, following the `redact-text` template in #149 under #147. That was not optional housekeeping: zeroclaw-labs/zeroclaw#9126 merged on 20 August, and a manifest that requests `config_read` without a schema stops being discovered or installed, so the previous revision of this branch would not install on a current host at all. Versions are 0.2.0 across `manifest.toml`, `Cargo.toml` and `Cargo.lock`, because the two allowlists in each plugin move from comma-separated strings to JSON arrays and that breaks an existing operator config.

Two relations stay in the guest code, since JSON Schema cannot express them between sibling properties: the warn threshold must sit above the critical one, and the durable nonce pair is set together or not at all. The guest-side unknown-key rejection is gone, replaced by two tests per plugin that read `manifest.toml` and assert the permission and the schema appear together and that every key the guest reads is declared. Both were mutation-tested.

**Synced with the standalone repository.** The branch had been parked since 5 August while the work continued in https://github.com/ZiBibro/solana-portfolio-sentinel. The three plugin directories are byte-identical between the two trees now, and every figure in the description was re-measured after the sync: 241 host tests, `fmt` and `clippy -D warnings` clean on the host target and on `wasm32-wasip2`, and a locked release build for `wasm32-wasip2`.

On timing and expectations: the 22 July guidance said registry merges happen after judging by maintainer invitation, and no invitation has been extended. This is ready rather than a request to jump the queue, and #147 is the reason it made sense to bring it up to the current contract now instead of waiting to be asked. Happy to rebase, split it per plugin, or put it back in draft if that suits the queue better.

ZiBibro:
Verified against a host built from current upstream `master` (`bd8c73f1`, which carries the typed-config commit `78a801ad`), rather than only against the pin these packages were developed on. That run found one real defect, now fixed in `86e4f76`.

**What the run did.** Built `zeroclaw` with `plugins-wasm-cranelift`, installed all three packages from a clean `--config-dir`, and loaded a `config.toml` written from the operator template.

```text
$ zeroclaw --config-dir "$C" plugin install .../plugins/lending-health
Plugin installed from .../plugins/lending-health
Seeded [[plugins.entries]] for 'zpi1_WyJsZW5kaW5nLWhlYWx0aCIsInRvb2wiLCJsZW5kaW5nLWhlYWx0aCJd'.

$ zeroclaw --config-dir "$C" plugin list
Installed plugins:
  stake-monitor v0.2.0 — ...
  stake-tx-build v0.2.0 — ...
  lending-health v0.2.0 — ...
```

All three are discovered and installed, and the host seeds a typed entry for each.

**The defect.** The migration moved the four list-shaped keys to JSON arrays and updated each package's key table, but `stake-monitor`'s README kept the 0.1.0 comma-separated form in the TOML block under that table. An operator pasting it gets a package that installs and then fails on every configured call. Fixed in `86e4f76`; the same shape in this project's reproduction guide was fixed alongside it.

**The rule the run confirmed**, worth stating plainly since it is easy to read the wrong way from the schema: operator storage stays `HashMap<String, String>`, so every value in `[plugins.entries.config]` is a quoted string, and the schema decides how each string is read. A bare TOML integer is refused before any plugin runs:

```text
Error: ... invalid type: integer `20`, expected a string
in `plugins.entries.config.timeout_secs`
```

and the resilient loader then discards the whole `plugins` section for that run with a single warning line. The corrected template loads with no warning at all.

Nothing else changed: no manifest, no version, no guest code. The branch is `86e4f76`, still 241 tests, and it remains ready for review whenever the registry queue reaches it.

ZiBibro:
Follow-up to the verification above: the same audit was widened from that one code path to every copyable example in the three packages, and it found two more defects in this branch. Both are fixed in `8313281`.

**The config entry was keyed by package name.** `stake-monitor`'s README showed

```toml
[[plugins.entries]]
name = "stake-monitor"
```

The host resolves entries by the plugin's instance key, so that entry is never consulted. The failure is silent in the way that costs the most time: `plugin install` succeeds, `config list` prints every value, the daemon starts, `plugin list` shows the package, and only the tool call fails, reporting the required key missing. The example now carries the key the installer seeds, with a comment saying where it comes from.

**The key-extraction line did not extract the key.** All three READMEs told the operator to run

```bash
key=$(zeroclaw plugin info <package>)
```

which returns the whole six-line record, so the `config set` path built from it cannot resolve. It is now piped through a grep for the key itself, checked against the real binary for all three packages:

```text
$ zeroclaw plugin info stake-monitor | grep -o 'zpi1_[A-Za-z0-9_-]*'
zpi1_WyJzdGFrZS1tb25pdG9yIiwidG9vbCIsInN0YWtlLW1vbml0b3IiXQ
```

**One behaviour change, in `stake-tx-build`.** The vote-allowlist refusal named the old comma-separated encoding, and that string is what a model relays to an operator who is already debugging their allowlist. It now names the quoted JSON array the current host requires. The manifest comment claiming an empty list differs from an omitted key is corrected alongside it: `txbuild.rs` maps both to the same empty allowlist, and the meaning is the safe one, so only `deactivate` can be built.

Package tests pass unchanged. Branch head is `8313281`.

ZiBibro:
One more result from driving these components through the real runtime, and this one is not about this PR: it looks like it affects every package in the registry, so it is yours to decide rather than mine to patch here.

A component built against `wit/v0` in this repository does not instantiate on a host built from current upstream `master`:

```text
component imports instance `zeroclaw:plugin/logging@0.1.0`, but a matching
implementation was not found in the linker: instance export `log-record` has the
wrong type: type mismatch with parameters: type mismatch for field action:
expected enum of 38 names, found 37 names
```

The host's `wit/v0/logging.wit` has gained a `memory-audit` variant on the log-action enum since `f009445` vendored this copy. Adding a case to an existing enum breaks previously compiled plugins — the contract [#9643](https://github.com/zeroclaw-labs/zeroclaw/issues/9643) asked for and [PR #9950](https://github.com/zeroclaw-labs/zeroclaw/pull/9950) wrote down — and this is that break, arriving with no diagnostic until instantiation.

Scope, as far as I can see it: all 34 packages here bind `path: "../../wit/v0"`, so the drift is shared rather than specific to mine. `tool.wit` has also gained `import secrets;` upstream, and the host now ships `config.wit` and `secrets.wit` that this tree does not carry; `channel.wit` differs by 87 lines. I have not checked whether those matter at instantiation the way the logging enum does — only the enum reproduced for me, because it is the interface every plugin imports.

Two things that cost me time and might save yours:

- `cargo` does not treat the WIT directory as a build input, since the bindings come from `wit_bindgen::generate!` rather than a build script. After changing a `.wit` file the crate has to be touched, or the rebuild silently reuses the old bindings and the same error survives it.
- After aligning only `logging.wit` and rebuilding, all three of my components instantiate, receive their typed config, and run. So for my packages the enum was the whole of it.

I have not touched `wit/v0` on this branch: it is shared, it is yours, and a package PR is the wrong place for it. Happy to open a separate issue with the reproduction if that is more useful than a comment here.

ZiBibro:
Correcting my own comment above, and updating the branch.

**The WIT drift I reported is already fixed here, and I missed that.** I diffed this repository's `wit/v0` from my branch, which was five commits behind `main`, and reported the gap as if it were open. It is not: #154 advanced `wit/v0` and pinned `cbf88733`, and that copy already carries the `memory-audit` variant a current host requires. My apologies for the noise — the reproduction was real, the conclusion about scope was not.

What I changed in response: this branch has merged `main`, so it is up to date and the components now build against the registry's own `wit/v0` rather than a copy I took from the host. Verified after the merge by driving each component through the real runtime: `lending-health` instantiates, receives its typed config, and returns a live Kamino position; the other two reach their RPC call. Component sizes are unchanged.

**Also worth flagging, since it decides whether you ever see a green tick here: CI has never run on this branch.** All eleven workflow runs on `feat/solana-portfolio-sentinel` sit at `action_required`, and `gh pr checks 63` reports `no checks reported`. That is the first-time-contributor approval gate, not a failure, but it means any "green" claim about this PR — including one I made in an earlier comment on my own repository — has no run behind it. Nothing to do on my side; it needs a maintainer to approve the workflow run.

One more correction to that same earlier comment: I wrote that the branch was ready for review while the PR body still opened by saying it was parked as a draft and should stay out of the queue. The body is corrected.

Branch head is `1647a62`.

ZiBibro:
Housekeeping note: this branch was force-pushed to replace an author email in its own commits with the GitHub noreply address. Content is unchanged — the tree at the new head is byte-identical to the tree at the old one — but every SHA on the branch moved, so `1647a62` in my previous comment no longer resolves. The current head is `4fab33f`, and the branch still carries `main` merged in, so it remains up to date with no commits behind.

Only commits authored on this branch were rewritten; the merged `main` history is untouched.

ZiBibro:
Judging closed on 21 August and the results are public, so this is a question about the merge path rather than a nudge for review.

The bounty listing said registry merges happen separately after judging, by maintainer invitation. If an invitation for this plugin family is planned, the branch is ready for it: it carries `main` with nothing behind, and the three manifests declare a closed Draft 2020-12 `config_schema` and consume typed `__config`. All three packages install from a clean config directory and load on a host built from current upstream `master`, with the run output in the comments above.

If the registry is not taking third-party submissions, that is a fine answer and I would rather hear it than leave this open. CONTRIBUTING points contributors to their own repositories, and the bounty exception ran to 21 August. Say the word and I will close this myself. The plugins live at https://github.com/ZiBibro/solana-portfolio-sentinel under the same gates in their own CI, so closing costs nothing.

One practical note either way. The validation workflow on this branch has never run: every attempt sits at `action_required`, which is the first-contribution approval gate. That one needs a maintainer click, and nothing I push from this side clears it.

PR #91, opened 2026-07-19
token-risk-check: T0 Solana mint safety verdict (RED/AMBER/GREEN) WASM tool plugin

rune-lynx:
## token-risk-check — T0 mint-safety verdict tool for Solana

One question, answered before your agent touches any token: **is this mint safe to hold?**

Give it any Solana mint address and it returns a compact RED/AMBER/GREEN verdict covering:

- live **mint / freeze authorities** (supply inflation, account freezing)
- **Token-2022 extension traps**: permanent delegate, transfer hooks, transfer fees, default-frozen, non-transferable (soulbound), and friends — parsed from the raw TLV
- **holder concentration** (top-holder share), with explicit disclosure when the RPC withholds it

Read-only **T0 custody tier**: no keys, no signing, no state. The worst a compromised call can do is read public chain data.

### Why this one

The bounty brief listed a token risk checker as the plugin "we'd like to exist most of all." It also happens to be the sharpest fit for the safety-first judging axis: the tool exists to give an agent grounds to *refuse* an action.

### Validated against reality (mainnet, 2026-07-19)

- **PYUSD → RED**: catches PayPal's **permanent delegate** (the canonical seizure-key trap this tool exists to find), parsing all 8 of its live Token-2022 extensions.
- **USDC → AMBER**: live mint/freeze authorities, correctly not RED.
- Free public RPCs block `getTokenLargestAccounts`; the plugin **discloses** the missing holder data instead of failing or guessing.

### Verified end-to-end inside the host

Run on a source-built v0.8.3 host (`--features agent-runtime,plugins-wasm-cranelift`, aarch64 phone): auto-discovery registers the tool next to the 50 built-ins, the *supervised* risk profile demands human approval before the first call, and on approval the component executes over the host's permission-gated `wasi:http` and returns the PYUSD RED verdict above.

Note for reviewers: shipped release binaries (≤ v0.8.3) exclude the `plugins-wasm` feature, so testing needs a feature-enabled build — exact steps are in the plugin README.

### Hard requirements checklist

- [x] Pure core (`spl.rs` TLV parser / `rpc.rs` codec / `risk.rs` verdict) + thin `#[cfg(target_family = "wasm")]` shim; `cdylib` + `rlib`
- [x] 19 host-run `cargo test` over mocked RPC — no network, no wasm toolchain needed
- [x] Structured logging via the WIT `logging` import, never stdout
- [x] `manifest.toml` with minimal permissions: `http_client`, `config_read`
- [x] README: custody-tier rationale, threat model, degradation contract, field notes
- [x] Repeatable **prompt-injection drill** (`examples/injection.rs`) — every adversarial `mint` argument fails closed before request-building; transcript in the README
- [x] Clean `wasm32-wasip2` release build (~340 KB)
- [x] MIT (dual MIT OR Apache-2.0)

### Output shaping

Verdict line + facts line + one line per reason: ~50–150 tokens in the agent's context, never the RPC's 40 KB.

---

Built for the Superteam Earn "Build Solana-native plugins for ZeroClaw" bounty. Happy to adjust anything to fit repo conventions — the registry entry is intentionally left to the publish workflow per the README.


rune-lynx:
Demo video (70s, covers this plugin and #92): https://github.com/rune-lynx/zeroclaw-plugins/releases/download/demo-v1/zeroclaw_plugins_demo.mp4 — assembled from real runs on a Samsung S25: local Qwen3.5-2B through a source-built v0.8.3 host (plugins-wasm-cranelift), the model choosing the tool unprompted, the supervised approval gate, and live mainnet execution. Verbatim tool outputs recovered from the runtime trace.

rune-lynx:
Update (8a52c93): while validating this plugin against a paid mainnet RPC I found the README was wrong about *why* holder-concentration data goes missing, so I corrected it and added verified transcripts of both paths.

The README previously attributed the gap to free public RPCs throttling `getTokenLargestAccounts`. That's real, but it is not the only cause — **every** RPC, paid ones included, rejects the call outright for tokens with very large holder counts:

```
USDC EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v
-32600: Too many accounts requested (10000000 pubkeys)
```

That's a protocol ceiling, not a rate limit, so no amount of endpoint quality fixes it. Worth knowing for anyone else building on this call.

Both paths now have transcripts in the README, same build, same endpoint:

- **PYUSD** — concentration available, and it changes the verdict: `top holder 56.6% / top10 83.4%`, adding `[RED] top holder owns 56.6% of supply (majority control)` on top of the permanent-delegate finding.
- **USDC** — RPC refuses; the plugin still delivers the authority/extension verdict and discloses the gap rather than estimating. The header line omits concentration figures entirely when they weren't measured, so an operator can always tell the difference.

No behaviour change — the degradation path already worked; this documents it honestly and proves it against a real endpoint. Happy to take feedback on any of it whenever you're reviewing.

rune-lynx:
**Update: both submissions now share a common Solana primitives crate.**

This PR now includes `plugins/solana-wasip2-core` — a dependency-light, wasm32-wasip2-friendly crate for the Solana primitives every plugin in this family needs (strict base58 pubkey decode, RFC 4648 base64, sha256 helpers, compact-u16, and an unsigned legacy-tx builder with per-index validation). 25 tests of its own.

For this plugin the refactor drops the external `bs58` and `base64` dependencies. One deliberate detail: the crate's decode errors echo raw input for debuggability, but this plugin's untrusted-argument edge (`validate_mint`) discards them in favour of its sanitized messages — prompt-injected bytes still never round-trip into output.

PR #92 adopts the same crate (its hand-rolled tx module collapses onto it, −172 lines). The two PRs vendor identical copies of the crate directory, so they merge cleanly in either order; if you'd rather take the core as its own PR first, happy to split it out.

PR #92, opened 2026-07-19
depin-attest: T1 hash-chained sensor attestation builder for DePIN devices (Track C)

rune-lynx:
## depin-attest — T1 hash-chained sensor attestations from a DePIN device

Companion to #91 (`token-risk-check`, Track D). This one is Track C: **a Raspberry Pi (or any box) running ZeroClaw is already a DePIN device — this plugin is the chain part.**

The agent takes a sensor reading with the host's hardware tools, calls `depin_attest {metric, value}`, and gets back an **unsigned** Solana transaction (base64) carrying a canonical, hash-chained attestation memo — plus a human-readable summary for the approval gate. A human approves and signs; the device address accumulates a publicly verifiable, tamper-evident chain of readings.

### Custody: T1, structurally

No keys in the plugin, and the built transaction contains a single memo instruction — **a transfer is not expressible in its output**. Full prompt-injection worst case: a bogus reading is proposed, the human sees it, and even a rubber-stamped approval costs only the network fee.

### The chain design

`{"v":1,"dev":…,"seq":42,"ts":…,"metric":"temp_c","val":"23.5","unit":"C","prev":"a1b2c3d4e5f60708"}`

- `seq` recovered from confirmed history in **one** RPC call (`getSignaturesForAddress` carries memos)
- `prev` commits to the on-chain signature of the previous attestation — gaps, forks, and reordering are visible to any verifier
- Blockhash expiry (the brief's trap #1) answered head-on: rebuilds are free and idempotent — same `seq` until one lands — and the output tells the agent to re-call after a lapsed approval; durable-nonce support sketched as next step

### Fail-closed surface (drill in the README)

Nine-attack injection drill, every one blocked with zero RPC calls: smuggled `recipient`/`amount` keys die at `deny_unknown_fields`; unlisted metrics die on the operator's allowlist; impossible readings die on operator bounds; payload-breaking strings die at numeric parsing; the device key cannot be supplied by the model at all.

### Verified end-to-end

Run inside a source-built v0.8.3 host (`plugins-wasm-cranelift`, aarch64 phone): discovery → approval gate → sandboxed `wasi:http` → live mainnet blockhash → attestation #1 built. Real transcript in the README.

### Hard requirements checklist

- [x] Pure core (`att.rs` / `tx.rs` / `rpc.rs`) + thin wasm shim; `cdylib` + `rlib`; transport-injected flow so host tests cover every branch
- [x] 21 host-run `cargo test` with canned RPC — no network, no wasm toolchain
- [x] Hand-rolled legacy tx wire format (no `solana-sdk`), golden-vector tested — `tx.rs` is deliberately shaped as a seed of the Track E shared core
- [x] Structured logging via the WIT `logging` import
- [x] `manifest.toml`: `http_client`, `config_read` only
- [x] README: custody tier, threat model, chain design, worked example, injection transcript, wasip2 field notes
- [x] Clean `wasm32-wasip2` release build (~370 KB)
- [x] MIT (dual MIT OR Apache-2.0)

---

Built for the Superteam Earn "Build Solana-native plugins for ZeroClaw" bounty, on real hardware (the demo device is a phone/Pi5 setup). Happy to adjust to repo conventions.


rune-lynx:
Demo video (70s, covers this plugin and #91): https://github.com/rune-lynx/zeroclaw-plugins/releases/download/demo-v1/zeroclaw_plugins_demo.mp4 — the depin_attest segment shows the on-phone model turning "the sensor reads 24.1 C" into attestation #1 with a live mainnet blockhash, through the approval gate, no keys held.

rune-lynx:
**Update: −172 lines — the hand-rolled tx module now lives in a shared, tested core crate.**

This PR now includes `plugins/solana-wasip2-core` (identical copy to the one in PR #91 — they merge cleanly in either order). The plugin's entire `tx.rs` — base58 pubkey decode, compact-u16, the unsigned attributable-memo transaction builder, base64, and the attestation chain-link hash — collapses onto the core, whose suite carries the exact byte-layout test plus 25 more. Drops the `bs58`, `base64`, and `sha2` dependencies here.

Net effect across the two submissions: one strict, wasm-component-friendly implementation of the wire-format primitives instead of two hand-rolled ones — the kind of shared foundation the plugin family needs if more Solana tools land after these. If you'd rather take the core as its own PR first, happy to split it out.

rune-lynx:
**Update: the worked example is now verified against a paid mainnet RPC — and it was wrong.**

PR #91 got this treatment already; this brings #92 in line.

Added `examples/live.rs`: a transport-injected harness that replays real `getSignaturesForAddress` + `getLatestBlockhash` responses through the actual pipeline. Same convention as the sibling `token-risk-check` plugin — fixtures come in via `curl`, so `cargo test` stays network-free.

Running it against a paid Helius mainnet endpoint showed the README's "real output" section was illustrative rather than real:

- it used the **all-ones System Program address** (`1111…1111`) as the stand-in device, which is a confusing choice for a doc that's about device identity;
- it quoted `lastValidBlockHeight` of **`296…`** — **stale by roughly 117 million blocks** against live mainnet (`413815626` at time of writing).

Both corrected, and the section now carries the exact commands to re-derive the transcript, so the numbers can be checked rather than trusted. I left a dated note in the README saying what was wrong and why, rather than quietly editing it.

17/17 plugin tests still green. Nothing about the plugin's behaviour changed — this is purely the difference between a doc that *looks* consistent and one that's been checked against the chain.

PR #112, opened 2026-07-21
Safe Hands: Solana transaction authorization suite (T0 guard + T1 builder + T1 Squads proposer)

Pratiikpy:
## What this is

Three wasm32-wasip2 tool components + a shared core that give a ZeroClaw agent a complete, safe path for moving Solana money:

- **solana-tx-authorize** (T0): decode → declared-intent binding → deterministic policy → simulate → ALLOW/REVIEW/DENY/UNKNOWN with reason codes. Holds nothing, builds nothing.
- **spl-transfer-build** (T1): unsigned SOL/SPL transfers (ATA-aware, memo) that refuse out-of-policy requests at build time.
- **squads-proposal-build** (T1): builds unsigned Squads v4 proposals — but only after **independently re-running the entire policy evaluation** from its own operator config. A caller-supplied ALLOW is never trusted (SH-TRUST-FORGED).

The agent proposes, a Squads multisig disposes. No component holds a key or signs anything.

## Prove it in one command (offline)

```bash
just prove-safety
```

All unit tests + a **20-fixture YAML attack arena** (amount/recipient/mint swaps, hidden transfers, authority handovers, unlimited delegates, durable nonces, fail-closed paths, and a forged caller-supplied ALLOW) + clippy -D warnings on host **and** wasm32-wasip2 + release builds of all three components.

## Already run live

- Full 3-tool flow in a real ZeroClaw agent (kimi-k3) on devnet, ending in an on-chain Squads proposal that was approved and executed (0.05 SOL moved) — signatures in EVIDENCE.md.
- Byte-exact encodings vs @solana/web3.js (message, ATA) and the official @sqds/multisig SDK (PDAs, both instruction encodings, Squads' inner TransactionMessage format).

## Notes for review

- Layout: libs/safe-hands-core + three plugins/* mirroring redact-text (pure core, thin wasm shim, crate-type = ["cdylib", "rlib"], committed Cargo.lock, vendored wit untouched).
- Permissions: only config_read + http_client.
- Happy to adjust layout/naming — opening early for feedback per the bounty guidance.


Pratiikpy:
Keeping this as a **draft** deliberately, and flagging why so it isn't read as an abandoned PR.

The Superteam Brasil listing was updated to say that registry PRs are not submissions and that registry merges happen separately after judging, with maintainers inviting the strongest implementation per plugin family. So this branch is not being pushed for review during the bounty window — the submission is a showcase post, and this PR is here as the code it links to.

The branch is current and the components are unchanged. Two things have been added since the last push, both aimed at making the work checkable rather than at moving this PR along:

- `just prove-safety` now also runs in public CI on a clean runner (host tests, the attack arena, receipt re-derivation, transparency-log replay, `clippy -D warnings` on host **and** `wasm32-wasip2`, release components, a capability check read from the compiled `.wasm` import tables, and execution in a real wasm runtime).
- Two host issues we hit while building are filed separately, with source references and repros: zeroclaw-labs/zeroclaw#9779 and zeroclaw-labs/zeroclaw#9780.

Happy to pick this up whenever maintainers want to look at it post-judging, and equally happy to rebase or split it per whatever the registry process ends up being.

PR #123, opened 2026-07-21
feat(plugins): PixZClaw — dual-rail BRL PIX + Solana Pay USDC invoicing (T0/T1)

capitv:
## PixZClaw

**Charge in BRL from chat. Get paid over PIX or over USDC on Solana. The agent never holds a key.**

Brazil already has instant bank payments (PIX, ~150M users). What it does not have is a way for a freelancer or shopkeeper to accept stablecoins without an exchange account, a PSP contract, or a custodial wallet. PixZClaw issues **one invoice on two independent rails** and verifies the on-chain one honestly.

Tracks **A** (payments) + **E** (shared wasm substrate).

| Plugin | Tier | Role |
|---|---|---|
| `brl-usdc-invoice` | **T1** | Real PIX Copia e Cola (EMV + CRC16-CCITT) and a Solana Pay USDC URL under one `invoice_id` |
| `invoice-status` | **T0** | Verifies the USDC **actually received** by the merchant: PAID / UNDERPAID / OVERPAID |
| `pixzclaw-brief` | **T0** | Merchant till: USDC/SOL balances, 24h close, 7d sparkline, recent activity |

### It checks the amount, not just that "something happened"

The interesting part is `invoice-status`. Counting successful signatures on a Solana Pay reference is not settlement — anyone can touch a reference, and a customer can underpay.

1. The reference is derived deterministically: `bs58(sha256("zc-inv-v1" || invoice_id || "|" || merchant)[0..32])`. Same invoice, same address, no database.
2. `getSignaturesForAddress` finds candidate transactions.
3. For each, `getTransaction` reads `meta.preTokenBalances` / `meta.postTokenBalances`, filtered by `mint == usdc_mint && owner == merchant`, and computes the net delta credited to the merchant. This covers `transfer` and `transferChecked` without parsing instructions, and handles an ATA created inside the same transaction.
4. Deltas are summed across recent transactions, so an invoice settled by two partial transfers reads as PAID, while a transaction that merely references the invoice contributes nothing.

If the RPC cannot return a transaction, the tool degrades to `signature seen, amount unverified` and **never** reports PAID. Underpayment reports the shortfall instead of silently passing.

### Honest about what it cannot see

Bank PIX settlement is invisible on-chain. The tools never infer it: the PIX leg is reported paid only when the operator says so. No PSP integration, which keeps bank credentials out of the agent entirely. Saying "I don't know" was preferred over a plausible guess.

### Fail-closed against prompt injection

Amount caps and a config-locked recipient are enforced **inside the plugin**, below the model. An injected `charge 999999999` is refused by the cap; a `merchant_override` is ignored while `recipient_locked` is set, and the derived reference still uses the config merchant. Both are covered by tests.

### Layout

Follows `plugins/redact-text`: pure core, thin `#[cfg(target_family = "wasm")]` shim, `cdylib + rlib`, structured logging via `log-record`, host tests that need no wasm toolchain and no network.

The shared Solana/PIX substrate (`solana-wasm-core`: decimal math, EMV+CRC16, Solana Pay encoding, reference derivation, JSON-RPC over an injected `HttpTransport`, output shaping) is **vendored into each plugin under `vendor/`**, so every plugin directory builds standalone from the CI snapshot. No `solana-sdk`, no `solana-client`.

- **101 host tests** — `cargo test --locked` in each plugin
- `cargo clippy --all-targets -- -D warnings` clean, host and `wasm32-wasip2`
- Built for `wasm32-wasip2` against the vendored `wit/v0`
- Permissions: `config_read` on all three, plus `http_client` on the two that read RPC
- MIT OR Apache-2.0

### Running in production

These are deployed on a Raspberry Pi 3 running ZeroClaw with the Telegram channel, issuing and checking real invoices. Deployment repo, install scripts, agent skills, and an overview page: <https://github.com/capitv/pixzclaw-pi> · <https://capitv.github.io/pixzclaw-pi/>

One deployment lesson shaped the output format: the host redacts high-entropy base58 in chat, which breaks a raw `solana:` URL. The invoice therefore delivers the Solana leg as a QR link (the QR still encodes the full pay URL) and keeps the PIX copy-paste code in a fenced block, so the whole message stays forwardable to the customer.


capitv:
## Validation run locally, since the workflows here need a maintainer to approve them

`validate.yml` has not run on this PR — first-time contributor workflows wait for
approval — so this is what the same checks produce on my machine, plus one real
failure they caught and I have now fixed.

### `fmt` — one genuine failure, fixed

`git diff --check` against the merge base was **failing** before this push. The
vendored core carried two markdown hard-break trailing spaces in its README and a
blank line at EOF in its `Cargo.toml`. The blank line was not hand-written: the
vendoring script cuts the crate's standalone `[workspace]` section, and that cut
left behind the blank line which had separated it, so every re-vendor reproduced
it. Fixed at the cause — the script now collapses trailing newlines to exactly
one — rather than by editing the generated file.

```
$ base=$(git merge-base origin/main HEAD); git diff --check "$base" -- plugins/
$ echo $?
0
```

`cargo fmt --all -- --check` passes on all three plugins. It still fails on nine
other plugins in the tree (`bluesky`, `mattermost`, `mochat`, `notion`, `reddit`,
`telegram`, `twitter`, `wechat`) — pre-existing debt that the transitional
baseline policy correctly warns about rather than failing, since this PR does not
touch their `.rs` files.

### `registry` — structure guard clean

```
$ for dir in plugins/*/; do for f in manifest.toml Cargo.toml Cargo.lock; do
    [ -f "$dir$f" ] || echo "MISSING $dir$f"; done; done
$ echo $?
0
```

`registry.json` is deliberately untouched — the three plugins are new names, so
there is no prior version whose bytes could shift under an existing identity.

### Components — tests and lints

| Plugin | `cargo test` | `cargo clippy --all-targets -- -D warnings` |
|---|---|---|
| `brl-usdc-invoice` | 20 passed, 0 failed | clean |
| `invoice-status` | 25 passed, 0 failed | clean |
| `pixzclaw-brief` | 7 passed, 0 failed | clean |
| `solana-wasm-core` (vendored into each) | 83 passed, 0 failed | clean |

135 tests total, no network in any of them: the RPC layer sits behind an injected
`HttpTransport`, so tests feed recorded JSON.

### What I could not run here, stated plainly

The `wasm32-wasip2` **link** step is blocked on this machine by a Windows
Smart App Control policy that refuses to execute `wasm-component-ld`
(`os error 4551`). Compilation for the target reaches the linker — so the code
type-checks and codegens for wasm — but I cannot produce the component locally
and will not claim otherwise. The components in the published release were built
on Linux from these exact sources.

### Version

Bumped to `0.3.1` so the three plugins name the bytes that are actually installed
and demonstrated in production, instead of a `0.3.0` that no longer corresponds
to any shipped artifact.

### If it helps review

The T0 verification is runnable without a ZeroClaw host at all:
[`examples/verify-live`](https://github.com/capitv/pixzclaw-pi/tree/main/examples/verify-live)
calls `invoice_status::status_tool::fetch_and_status` — the same function the
component calls — and swaps only the injected transport (`waki` over `wasi:http`
in the plugin, `curl` there). It prints the address it is about to read and an
explorer link to it before it prints a verdict.

Happy to rebase or split this if any of it is easier to review in pieces.

Issue #147, opened 2026-07-30
config_schema is now required for every plugin that requests config_read (all 31 packages)


JordanTheJet:
## What is changing

ZeroClaw is adding typed, schema-validated plugin config (host PR [zeroclaw-labs/zeroclaw#9126](https://github.com/zeroclaw-labs/zeroclaw/pull/9126)). Once it lands, **a manifest that requests `config_read` without declaring `config_schema` is no longer discovered or installed.** The two are a biconditional: declaring a schema without requesting the permission is equally invalid.

This ships without a compatibility shim or grace period. Plugins are a pre-1.0 experimental surface, and an untyped fallback would have to hand a guest values the host cannot type, name, or bound, which is the hole the feature closes. Packages that do not migrate stop being discovered; nothing is silently downgraded.

Every package in this repository requests `config_read` and none declares `config_schema` today, so **all 31 need migration.**

## Why the host needs the schema

Operator values are stored as a secret-marked string map, encrypted at rest, and the guest is untrusted code. Before starting a guest the host must know which keys the package may receive and what type each value is. The WIT worlds are fixed and shared across all plugins, so per-package config types cannot live in the ABI. The manifest is the only place the contract can be declared, and `additionalProperties = false` plus an explicit `properties` map is what makes the `config_read` grant enumerable.

## What each package needs

Full recipe with examples, limits, and error messages: **[Migrating to typed plugin config](https://github.com/zeroclaw-labs/zeroclaw/blob/master/docs/book/src/plugins/migrating-to-typed-config.md)**.

Short version, per package:

1. Add a closed Draft 2020-12 `[config_schema]` to `manifest.toml` covering exactly the keys the plugin reads. Root must be `type = "object"` with `additionalProperties = false`; every top-level property needs one explicit type from `string`, `boolean`, `integer`, `number`, `array`, `object`.
2. Mark credentials `required` so a withheld grant fails closed instead of starting half-configured. Optional keys receive `{}` when the grant is withheld, so give them guest-side defaults.
3. Update the guest to deserialize the injected typed JSON once instead of parsing strings. Booleans, numbers, arrays, and objects arrive as real JSON values.
4. Rebuild the component.
5. Re-sign, if the package is signed: `config_schema` is covered by the manifest signature.

```toml
permissions = ["config_read"]

[config_schema]
"$schema" = "https://json-schema.org/draft/2020-12/schema"
type = "object"
required = ["bot_token"]
additionalProperties = false

[...]

PR #149, opened 2026-07-31
feat(redact-text): declare config_schema and consume typed __config (template for #147)

JordanTheJet:
## Summary

Template migration for #147: the canonical reference plugin (redact-text) gains its `config_schema` and consumes typed `__config`, ahead of the host enforcement landing in zeroclaw-labs/zeroclaw#9126. The other 30 packages should copy this PR's shape.

- **Schema**: closed Draft 2020-12, all-optional (`replacement` string `minLength=1`, `redact_emails` boolean, `patterns` array of non-empty strings), `additionalProperties = false`. Validated against the real host validator (`validate_manifest_config` / `resolve_plugin_config` from the zeroclaw#9126 head) with per-constraint rejection probes and a withheld-grant `{}` acceptance check; an un-migrated manifest fails the same harness, so the check is not vacuous.
- **Guest change is REQUIRED, and skipping it is a hard break**: with `__config` typed, the old `HashMap<String, String>` field fails the entire `ExecuteArgs` parse and every call returns "invalid arguments". `RedactConfig` now derives `Deserialize` with serde defaults; `from_json` replaces `from_section`; JSON null (host injected nothing) yields defaults; malformed config errors, never silently defaults.
- **Config errors cannot leak config values**: `from_json` returns a pre-sanitized error string because `serde_json::Error`'s `Display` quotes the offending value, and this plugin's `patterns` are exactly the strings an operator wants scrubbed. Mutation-checked: the naive version leaks the value verbatim into the `ToolResult`.
- **Version 0.3.0 in three files** (`manifest.toml`, `Cargo.toml`, `Cargo.lock` — CI runs `--locked`). `registry.json` untouched; `build-registry.py --check-metadata` classifies the new version as `pending unpublished source`, and release CI publishes it.
- README updated from the now-false comma-separated `patterns` documentation to the JSON-array encoding and instance-key flow.

## Checks

```text
cargo test --locked                               14 passed (baseline 7)
cargo clippy --locked --all-targets -- -D warnings clean (host + wasm32-wasip2)
cargo build --locked --target wasm32-wasip2 --release   ok (186,999 bytes)
cargo fmt --check                                 clean
tools/build-registry.py --check-metadata          11 matched + redact-text@0.3.0 pending
tools/ci tests                                    36 OK on a pristine tree
```

Mutation evidence: six guards disabled individually, each failing exactly its covering test (including the leak guard and the empty-pattern filter, whose removal rewrites entire inputs into replacement markers); all files restored byte-identical.

## Template notes for the remaining 30 (also posted to #147)

1. Version bump is three files; the lockfile is the one that gets forgotten.
2. Assume a guest change is needed; grep for `HashMap<String, String>`, `eq_ignore_ascii_case`, `.split(',')`.
3. Never let a config error echo the config value.
4. Keep guest-side normalization even where the schema duplicates it.
5. Decide `required` per key deliberately: credential-bearing channel plugins should fail closed, unlike this one.
6. Do not add `deny_unknown_fields` guest-side.


JordanTheJet:
Acceptance-tested end to end against a host binary built from the enforcing typed-config line. **This PR's own content is correct and complete**; it exposed a repository-level defect that is fixed separately in #154, and it is proven working on top of that fix.

**What passed on this PR alone:** 14 unit tests, clippy, a reproducible `wasm32-wasip2` release build, `build-registry.py --check-metadata` (11 indexed entries match, `redact-text@0.3.0` correctly classified pending), schema-gated install admission, `zpi1_…` key seeding with the printed command, and all three typed `config set` values applied.

**What failed, and why it is not this PR:** execution. The component could not instantiate on the host:

```text
instance export `log-record` has the wrong type: type mismatch for field
action: expected enum of 38 names, found 37 names
```

The vendored `wit/v0` pin points at an unmerged host branch commit that predates the `memory-audit` logging case released in v0.8.4. That affects **every** artifact this repository builds, including the published `0.2.0` entries — not just this migration. #154 splits the tree: `wit/v0` now tracks released hosts byte-identically, and the transport drafts move to `wit/unstable`.

**With #154 applied, rebuilt from this branch's source unchanged:**

```text
granted:  "mail a@b.com about project-zeus today" -> "mail [MASKED] about [MASKED] today"
withheld: "mail a@b.com about project-zeus today" -> "mail [REDACTED] about project-zeus today"
```

The granted run proves the operator's string, boolean, and array all arrived as typed JSON through discovery, admission, materialization, and injection. The withheld run proves the config jail: defaults apply and the operator's pattern is absent. An ill-typed `redact_emails = "yes"` is rejected host-side before the guest runs.

So the template does exactly what it claims. Merge order is #154, then this. No changes are needed here; the two branches touch disjoint files.

PR #154, opened 2026-08-18
fix(wit): pin wit/v0 to released hosts and move drafts to wit/unstable

JordanTheJet:
## Summary

Every artifact built from this repository — including the published `0.2.0` registry entries — fails to instantiate on released ZeroClaw hosts. Found while acceptance-testing #149 end to end against a v0.8.4-line host:

```text
component imports instance `zeroclaw:plugin/logging@0.1.0`, but a matching
implementation was not found in the linker: instance export `log-record` has
the wrong type: type mismatch for field action: expected enum of 38 names,
found 37 names
```

**Root cause.** `wit/UPSTREAM_REF` pinned `e112ce6b5`, a host commit reachable only from an unmerged branch. It predates host PR #9258, which added the `memory-audit` case to `plugin-action` **and shipped in released v0.8.4**, while carrying `sockets.wit`, `ws-client.wit`, and webhook-ingress additions that `master` never merged. Component enum arity must match exactly in both directions, so the missing case alone is fatal. Because no host commit carries both sets, advancing the pin could not satisfy the byte-identical WIT drift gate — the tree had to be split.

**The split.**

- **`wit/v0`** is now byte-identical to released **v0.8.4** and current `master` (verified identical to each other, so one target serves both), with the pin advanced to `cbf88733605e2827ee491758d29eaa4486b2940c`. The drift gate passes against it.
- **`wit/unstable`** is the pre-sync tree: the transport and webhook drafts, plus the `memory-audit` case released hosts require. The thirty plugins that import those interfaces bind here and stay `registry = false` until the host ships them. `README-UNSTABLE.md` states the contract.
- **`redact-text`** binds `wit/v0` — the first plugin on the released ABI.

## Testing

Rebuilt and executed against a host binary built from the enforcing typed-config line:

```text
cargo build --target wasm32-wasip2 --release   redact-text ok (187012 bytes), telegram ok (unstable path)
zeroclaw plugin install <pkg>                  admitted; seeded zpi1_… and printed the config-set command
zeroclaw config set …config.{replacement,redact_emails,patterns}   all three typed values applied
```

End-to-end execution through real discovery, admission, typed materialization, and the guest:

```text
granted:  "mail a@b.com about project-zeus today" -> "mail [MASKED] about [MASKED] today"
withheld: "mail a@b.com about project-zeus today" -> "mail [REDACTED] about project-zeus today"
```

The granted run proves the operator's string, boolean, and array all reached the guest as typed JSON; the withheld run proves the config jail (defaults apply, the operator's pattern is not). An ill-typed `redact_emails = "yes"` is rejected host-side before the guest runs. Before this change the identical package failed at the linker; after it, every step passes.

`plugins/redact-text` unit tests (14), clippy, and `tools/build-registry.py --check-metadata` are green.

## Scope and consequences

- **Breaking for the fleet, and already broken:** the published `0.2.0` entries do not load on released hosts today. They need rebuilding and republishing against `wit/v0`; that wave is tracked with the `config_schema` migration in #147, and each plugin's move from `wit/unstable` to `wit/v0` belongs in its own PR alongside its schema work.
- **Socket, WebSocket, and webhook plugins are unaffected in kind:** they could not run on released hosts before this change either, and remain `registry = false`. What changes is that the repository now says so structurally instead of implying a single supported ABI.
- No plugin logic is modified. The thirty repointed files change only the `wit_bindgen::generate!` path.

Depends on nothing. #149 rebases onto this and is proven working on top of it.


JordanTheJet:
Two findings from continued testing that reviewers should have before judging this PR. Neither changes what this branch does; both sharpen what it does **not** do.

**1. `wit/unstable` is not a path back to released hosts.** I probed a published HTTP-only plugin (telegram) built from this branch's `wit/unstable` against a released-line host:

```text
failed to convert function to given type: type mismatch with results:
expected flags type with 22 names, found 23 names
```

That is `channel-capabilities`: the unstable `channel.wit` carries the `webhook-ingress` flag (23), released hosts have 22. So repointing a plugin to `wit/unstable` keeps it unloadable on released hosts — and it cannot simply be repointed to `wit/v0` either, because the same plugins *export* `webhook-path` / `parse-webhook` returning `webhook-rejection`, a type the released world does not define.

**Consequence:** moving a plugin onto the released ABI is a **source migration** (drop or gate the webhook exports), not a bindgen-path change. This branch therefore fixes `redact-text` only — proven end to end — and leaves the other thirty exactly as loadable as they are on `main` today: not at all. That is the status quo, not a regression, and it is why each plugin's move belongs in its own PR alongside its `config_schema` work in #147.

(Probe caveat for whoever repeats it: grant `http_client` in the scope, or instantiation fails earlier on a `wasi:http` linker error that masks this one.)

**2. The release-planning trigger needs a decision before this can merge.** `tools/ci/plan_matrix.py` sets `release_all` when any changed path starts with `wit/v0/`. This branch changes five paths there, so **all 31 plugins become release inputs**, and `build-registry.py` then rejects each one whose manifest version is already published: `changed release input reuses immutable package identity`. Twelve names collide (`bluesky gitea matrix mattermost mochat notion redact-text reddit telegram twitter wechat wecom`), eleven after #149 bumps `redact-text`.

Mass-bumping those eleven would republish artifacts that finding 1 shows are still unloadable — publishing new versions of known-broken components. The alternative is to make a `wit/<tree>` change mark as release inputs only the plugins that bind that tree, which is the accurate model now that two trees exist and would leave this branch releasing `redact-text` alone. That is a repository-policy call, so I have not made it here.

Everything else on the branch is green: all 31 plugins pass tests, host clippy, wasm clippy, and `cargo fmt --check` under the repo's own `validate_components.sh` (stricter than CI, since it treats every plugin as strict).

JordanTheJet:
Correcting my earlier comment: **the release-identity concern I raised does not apply, and this branch needs no version bumps.** I raised it from reading `plan_matrix.py`'s `release_all` trigger rather than from observed behavior, which was the wrong basis. Required CI is now green (10/10), and the packager's own output settles it:

```text
PLANNED_MATRIX_JSON: … "release_plugins":[] … (all four shards)
  reused immutable telegram v0.2.0  sha256=953ed3c24ad0…
  reused immutable redact-text v0.2.0  sha256=65176a6eeb33…
verified exact publication set with 0 new archives
```

Every published name takes the `reused immutable` path, so nothing collides. Checking the publish side too: `publish.yml` never invokes `plan_matrix`, so `release_plugins` is not populated there either; it re-verifies the already-validated artifact with `--check-publication` and `--check-metadata`. There is therefore no live code path on this change that demands a bump.

That matters beyond bookkeeping, because bumping would have been the harmful option: the probe in my previous comment shows the thirty plugins on `wit/unstable` still cannot instantiate on released hosts, so republishing them at new versions would have shipped new versions of known-broken components. Leaving them at `0.2.0` is correct until each one's source migration lands.

What this branch actually does, restated with that correction: it makes `wit/v0` byte-identical to released hosts and advances the pin, moves the unmerged drafts to `wit/unstable`, teaches the component validator to snapshot both trees (without that, any plugin binding the second tree cannot compile in CI), and pays down pre-existing `fmt` and clippy debt in the eight plugins the repoint pulled into the validated set. `redact-text` is the one plugin on the released ABI, and it is proven end to end: install, schema-gated admission, typed config materialization, execution, and the config jail.

JordanTheJet:
Second correction, and this one supersedes the first. I now have the mechanism, measured rather than reasoned.

**CI computes changed plugins per push, not across the whole PR.** That single fact explains both of my earlier, contradictory readings:

- At `4515242` the push contained only `tools/ci/validate_components.sh`, so no plugin was a changed release input, the matrix carried `release_plugins:[]`, and the packager happily reported `verified exact publication set with 0 new archives`. I read that as "the blocker does not exist." It only meant "that push touched no plugins."
- At `042cf51` the push contained the clippy paydown across eight plugins, so exactly those eight became changed release inputs, and the packager failed each one: `changed release input reuses immutable package identity bluesky@0.2.0`, and the same for mattermost, mochat, notion, reddit, telegram, twitter, and wechat. Eight failures for the eight plugins that commit touched, which is the mechanism stating itself plainly.

So the rule is: **touching a published plugin's source makes it a release input and demands a version bump.** The `wit/v0` change alone never did.

**Consequence, and a reversal of my own recommendation.** I previously argued for keeping the lint paydown in this PR. That was wrong, and the evidence is what changed my mind: including it forces version bumps on eight published plugins, and bumping them republishes components that my earlier probe shows still cannot instantiate on released hosts. Paying down lint debt is not worth shipping new versions of known-broken artifacts.

The clippy commit is therefore dropped from this branch and preserved at `backup/zcp-lint-debt-paydown`. Each plugin's lint fix belongs in that plugin's own migration PR, where a version bump is legitimate because the artifact actually becomes loadable.

This branch is back to three commits: the WIT split, the `cargo fmt` paydown, and the validator fix that lets a plugin build against the tree it declares. It touches no published plugin's source, so nothing collides.

